Issue #18, opened 2015-02-18
USB Raw devices


t-kopp:
USB Raw devices don't seem to be supported at the moment. Is this just some bug I don't see or is there something missing? If the latter is the case I'll iron out a couple issues with my locally working solution and submit it as a PR

Edit: I've seen a couple devices out there which simply had 1 Bulk in and one out, which is what I would target to add

hgrecco:
They are not supported right now. It is a bug: lack of time of the developers :-D
It would be great if you can submit a PR. Remeber to register the Session class that you create using the following decorator:
```python
@Session.register(constants.InterfaceType.usb, 'RAW')
```

t-kopp:
yes, currently did it dirty by replacing INSTR support :S will clean that up and come back to you

hgrecco:
Mereged in #19

PR #19, opened 2015-02-18
Adding USB Raw device support

t-kopp:
This adds support for USB Raw devices (issue #18 , unfortunately I don't have any hardware to test INSTR devices. So a quick check can't hurt.


hgrecco:
Looks like a good start. Still we would need to implement the attributes of USB INSTR and USB RAW. Why don't you create a `USBSession` with all the common code. And then `USBInstrSession` and ``USBRawSession` deriving from it? It might be easier to maintain in the future.

hgrecco:
Merged. I just removed abc because I think is not necessary for a single method in classes that we control. Thanks a lot.

PR #27, opened 2015-03-23
More GPIB stuff

bessman:
I wanted to implement event disable/discard before opening this pull request, but I don't know when I'll have time to look into that so in case you want to do a release soon, here are some more GPIB things.


hgrecco:
Merged. Thansk

PR #32, opened 2015-05-05
gpib.read() is already synchronous.

bessman:
Turns out reading one byte at a time can cause timeout problems with certain interfaces. A NI-GPIB-USB-B adapter, for example, takes roughly two seconds to respond to *any* read operation, regardless of how many bytes are read. Reading 50 bytes one at a time would therefore take 100 seconds, unless it timed out first.

Luckily, linux-gpib reads synchronously anyway so there is no need to read one byte at a time.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,8 +3,14 @@ pyvisa-py is written and maintained by Hernan E. Grecco <hernan.grecco@gmail.com
 
 Other contributors, listed alphabetically, are:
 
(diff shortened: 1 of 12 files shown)


hgrecco:
By the way, this is something that was bugging me about all implementations. Is there any way **not** to read 1 byte at a time but still conserve all the read termination capabilities that NI-VISA provides?

Issue #38, opened 2015-05-20
Support for TCPIP SOCKET


MatthieuDartiailh:
It would be nice to have support for TCPIP SOCKET protocol, as TCPIP is the only protocol that should always work. Could you give some pointers about what would need to be adapted from the INSTR case to support the SOCKET case ?

nktsaini:
I too am looking for a way to use a raw socket connection using the pyvisa-py. Is any such thing forthcoming in the near future?

nktsaini:
Hey Matthieu.

I was just digging around trying to send SCPI commands over the raw socket connection since my instrument understands SCPI only. I ran the following script:

```
import socket
sock = socket.socket(sock.AF_INET, sock.SOCK_STREAM)
sock.sento("DISP::TEXT::STAT ON", ("10.66.22.150", 1349))
```

I don't get any error messages but my instrument doesn't respond either. So I think the information is being sent but the instrument isn't understanding it. If you have an instrument, could you confirm this?

MatthieuDartiailh:
Your message is lacking a termination character. You will have to check the doc to see what it expects (\r\n is a common combination).

nktsaini:
You're right. 

Also, I think the sendto() module sends one character at a time and hence needs to be used in a loop. If I run the following script, I am able to communicate with the instrument successfully.

```
import socket
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.sendall("[Your random SCPI command]\n")
```

hgrecco:
I will add support for SOCKET if you are willing to test it.

MatthieuDartiailh:
Great !

nktsaini:
Yes, let us know how do we start.

hgrecco:
I have pushed a commit in this direction.

thliebig:
I have fixed some issues and got a first working SOCKET connection, thanks

nktsaini:
I am running this script:

```
import visa
rm = visa.ResourceManager('@py')
dmm = rm.open_resource('TCPIP::10.66.22.150::1394::SOCKET')
```

and I get the error message saying
```
Exception: could not create socket: an integer is required
```
What am I doing wrong here?

thliebig:
I have fixed this and more in my commits ... please try these

nktsaini:
I ran the following script:
```
import visa
rm = visa.ResourceManager('@py')
[...]

PR #42, opened 2015-06-02
PR for issue #38

thliebig:


hgrecco:
Thanks a lot!

thliebig:
@hgrecco How do you test use your development version of pyvisa-py? I had to run `python setup.py install` after every change to test it, otherwise pyvisa would always use the installed version of pyvisa-py only...

hgrecco:
I have an isolated virtualenv with a symlink in to my development folder.

MatthieuDartiailh:
You can use ```python setup.py develop```. When doing so only a link to your development folder is added to pythonpath.

PR #45, opened 2015-06-15
Modify serial.timeout to send seconds to pyserial.

djanderson:
pyserial expects a timeout in seconds, but PyVISA's convention is to use milliseconds.


hgrecco:
Thanks for the PR but this was already fixed in a different one.

PR #50, opened 2015-07-20
VXI11 protocol performance enhancement

alexforencich:
Extending a bytearray is much faster than doing string manipulation for large transfers.


hgrecco:
Merged! Thanks a lot.

Issue #48, opened 2015-07-06
Very slow visalib.read() with instruments connected by USB


ayrodriguez:
The visalib.read() function takes about 6 seconds to return, whereas with usbtmc it's instantaneous.

pyvisa version 1.8.dev0
pyvisa-py version 0.2.dev2

hgrecco:
Sorry but I missed this, can you provide more info.

herm:
I'm having the same issue.
The problem is located here:
https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/usb.py#L98

USB communication never checks for the termination character but instead always uses the timeout.

hgrecco:
Please submit a PR, I will be happy to review it and merge it during the weekend.

herm:
I'm not sure how to fix it correctly. 
``lambda current: term_char_en and current.endswith(chr(term_char))`` 
works, but I think term_char_en should default to True for TMC but I don't know what the best place would be to set it.

herm:
This should be fixed now.

hgrecco:
@ayrodriguez Please confirm that this is fixed for you and close it! Thanks for the report and Thanks @herm  for the fix!

hgrecco:
I am closing this for now. Feel free to reopen or open another if necessary.

PR #63, opened 2015-10-12
USB TMC fixes

herm:
These patches fix the slow responses from USB devices (#48).
The USB code always returned a complete message, but the code in session.py only expected a single byte. Now the code handles the complete message, because returning single bytes from USB would be very inefficient. 

Tested with python 2.7 and 3.4 using a Keysight 34461A multimeter.


hgrecco:
I think this deserves a new version. Lets try it for a week and if everything is ok we push 0.3 Thanks!!!

brandonrwin:
Any possibility of getting this into PyPI? Version 0.2 is still listed, with this bug present.

PR #64, opened 2015-10-27
Pass timeout to USB-TMC code.

herm:
Without these changes the timeout is ignored for USB-TMC devices.


hgrecco:
Thanks for the PR. I think it would be better to do it directly when changing timeout as it is done in serial Take a look at: https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/serial.py

herm:
Updated.

hgrecco:
Thanks a lot, and sorry for the REALLY long delay.

PR #59, opened 2015-09-24
Did not work on python 3, without the encode

DavidFabijan:
Just a minor fix for GPIB communication on python 3.

PR #66, opened 2015-10-27
Revert 84fe24d4.

herm:
Revert 84fe24d4.
With this code transfering binary data is not possible. Tested on both python 2.7 and python 3.4.

### DIFF
--- a/pyvisa-py/sessions.py
+++ b/pyvisa-py/sessions.py
@@ -320,7 +320,7 @@ class Session(compat.with_metaclass(abc.ABCMeta)):
         out = b''
         while True:
             try:
-                current = reader().encode('ascii')
+                current = reader()
             except timeout_exception:
                 return out, constants.StatusCode.error_timeout
 


MatthieuDartiailh:
This does look a real issue and according to the discussion in the commit, the commit can be safely reverted. I will trigger bors next Monday if nobody is against.

MatthieuDartiailh:
Actually this seems to raise quite a lot of issue see related fixes in #67 #86 #70.

skrchnavy:
This is revert of change in #59 and is duplicate of #86 . And #70 wouldn't work if encode called.

Function is called from 3 files only:
* pyvisa-py\gpib.py
* pyvisa-py\serial.py
* pyvisa-py\usb.py

I propose to accept this and caller shall send proper reader (with encode if necessary).
For example in usb.py, this function is called:
```
 return self._read(lambda: self.interface.read(count),
                          count,
```
So when not working for usb, it shall be changed to:
```
 return self._read(lambda: self.interface.read(count).encode('ascii'),
                          count,
```
Similar code is in serial.py.

MatthieuDartiailh:
@skrchnavy could you make a new PR:
- reverting the encode
- documenting the fact that the reader is expected to return bytes/bytearray
- use a bytearray like was done in https://github.com/pyvisa/pyvisa/pull/282 to speed up the reading and convert to bytes before returning. 
- ensure that all current reader do properly return bytes and not string

I would prefer to see every issues being fixed at once and hopefully once and for all.

skrchnavy:
OK, I will test it with both python 2.7 and 3.6 to assure working common way

MatthieuDartiailh:
Closed by #119

PR #67, opened 2015-11-03
Add serial poll function.

herm:


MatthieuDartiailh:
Could somebody second that change ? I cannot test it.

MatthieuDartiailh:
This is reasonable and should have been merged years ago.

bors r+

PR #68, opened 2015-11-03
Catch GPIB errors while enumerating devices.

herm:
This avoids breaking the device enumeration for other device classes when there is no GPIB interface available but linux-gpib is installed.


hgrecco:
Merged. Thanks!

PR #70, opened 2015-11-16
Resolved the problems with binary gpib reads in python3

DavidFabijan:
The basic problem seems to be in the python binding for linux-gpib whihc return a string instead of a bytes array. The bottom solution works, but is in my opinion not very pretty. Multi-byte reading could/should probably be solved in the gpib module itself.


MatthieuDartiailh:
Closed by #119

PR #76, opened 2015-12-29
Fix serial call parameter name.

fake-name:
Fix for https://github.com/hgrecco/pyvisa-py/issues/75

### DIFF
--- a/pyvisa-py/serial.py
+++ b/pyvisa-py/serial.py
@@ -65,7 +65,7 @@ class SerialSession(Session):
         else:
             cls = Serial
 
-        self.interface = cls(port=self.parsed.board, timeout=2000, writeTimeout=2000)
+        self.interface = cls(port=self.parsed.board, timeout=2000, write_timeout=2000)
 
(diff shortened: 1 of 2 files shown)


fake-name:
This also fixes another bug with the serial data decoding that I think was probably because the code was never tested under python3k.

MatthieuDartiailh:
@fake-name I fail to see why you decode text_type, should it be encode ? Furthermore six is not a dependency, so we would need to add it.

fake-name:
Yeah, I'm not sure how that worked. It did work, at the time?

Anyways, I no longer have access to the hardware I used to `pyvisa-py`, so I can't really help much. It's been a year and a half!

skrchnavy:
#112 is follower of intended fix. Change in  session.py is related to #66 / #70.

I propose to close this PR without merging.

MatthieuDartiailh:
The issue with timeout parameter is addressed in #112 and the remaining one will be addressed in another PR.

PR #81, opened 2016-04-25
Speed up large transfers over GPIB

Duncanla:
Reading one byte at a time from linux-gpib can be very slow. This change
passes the read count directly to linux-gpib, making large transfers
much faster.


hgrecco:
Merged. Thanks!

Issue #83, opened 2016-06-23
Query over TCPIP that should timeout instead returns a blank string?


joshsleeper:
Although it should be raising a timeout error, all I get back is the empty string and PyVisa continues on as if everything is alright.
I've traced it all the way down to `rpc.Client.make_call()` ([link](https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/protocols/rpc.py#L238)) and the query is definitely returning error code `15`, which represents a timeout according to the `vxi11.ErrorCodes` enum ([link](https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/protocols/vxi11.py#L56)).

It _looks_ like the only possible places that the error would be raised are in `highlevel.PyVisaLibrary._return_handler()` ([link](https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/highlevel.py#L155)) and `highlevel.PyVisaLibrary.list_resources()` ([link](https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/highlevel.py#L238)), and neither of those seem to be hit for some reason.

I'm having a hard time tracking this down, but it's forcing us to roll back to installing a more commonly used VISA lib, and we dislike that very much. ;)

Any thoughts as to why this might be happening?

joshsleeper:
Hey @hgrecco,

I've researched a bit more, and it looks like `highlevel.PyVisaLibrary._return_handler()`([link](https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/highlevel.py#L155)) might never actually be called, which would explain why read and write timeouts aren't being handled properly. In fact, a comment from the docstring explicitly states `TODO: THIS IS JUST COPIED PASTED FROM NIVisaLibrary.`

Debugging my way through a query that times out using the standard NI Visa (and thus `pyvisa.ctwrapper.highlevel.NIVisaLibrary`), I _believe_ I discovered a handful of things. 

[...]

PR #85, opened 2016-07-07
TCPIP error handling fixes

joshsleeper:
Added error handling for `VisaIOError`s, raised with the appropriate error code.
Added error handling for `socket.timeout`, raised when the TCPIP-connected instrument stops responding. This could be due to a network issue or app crash, but it covers the case where the socket is left open but no communication is happening.
Minor code cleanup, removing unnecessary `str.split()` functions for example.

### DIFF
--- a/pyvisa-py/protocols/rpc.py
+++ b/pyvisa-py/protocols/rpc.py
@@ -28,6 +28,7 @@ import sys
 import enum
 import xdrlib
 import socket
+import select
 
 from pyvisa.compat import struct
 
@@ -254,12 +255,12 @@ class Client(object):
 
     def start_call(self, proc):
         # Don't override this
-        self.lastxid = xid = self.lastxid + 1
+        self.lastxid += 1
         cred = self.mkcred()
         verf = self.mkverf()
         p = self.packer
         p.reset()
-        p.pack_callheader(xid, self.prog, self.vers, proc, cred, verf)
+        p.pack_callheader(self.lastxid, self.prog, self.vers, proc, cred, verf)
 
     def do_call(self):
         # This MUST be overridden
@@ -309,8 +310,9 @@ def recvfrag(sock):
         buf = sock.recv(n)
         if not buf:
             raise EOFError
-        n = n - len(buf)
-        frag = frag + buf
+        n -= len(buf)
+        frag += buf
+
     return last, frag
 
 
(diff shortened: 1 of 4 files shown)


pyvisa-py/protocols/rpc.py
```diff
@@ -429,11 +456,7 @@ def make_call(self, proc, args, pack_func, unpack_func):
             pack_func(args)
         call = self.packer.get_buf()
         self.sock.sendto(call, (self.host, self.port))
-        try:
```
MatthieuDartiailh: Any idea why it was done this way before ?

MatthieuDartiailh:
This looks good to me. @hgrecco could you give a look ?

pyvisa-py/protocols/rpc.py
joshsleeper: First off, thanks for much for giving this a look over @MatthieuDartiailh. :+1:

As for why it was done this way originally, I can't say for sure, but I can think of two possible reasons:

1. To shorten the call (`select(r, w, x)` instead of `select.select(r, w, x)`)
2. To gracefully allow for the possibility that `select` could fail to be imported and continue with the warning that the socket hanging is a very real possibility. That said, I can't see why this would _ever_ happen here. `select.select()` has been in the Python core modules since 2.5, and neither pyvisa nor pyvisa-py claim to formally support anything below Python 2.6.

Since I ended up needing it in `do_call()` and I tend to prefer the absolute call style (especially if they have generic names), I figured it was justified to just import at the module level and use it the same everywhere.

hgrecco:
I was disconnected from this for a couple of weeks. It looks good to my and a good adhesion. I will merge it on the weekend and release a new version early next week. Thanks a lot.

joshsleeper:
That's great to hear @hgrecco, thanks so much!

pyvisa-py is a godsend for me and my team, so I'm more than happy to help make it better where I can haha.

MatthieuDartiailh:
ping @hgrecco

joshsleeper:
ping 2 @hgrecco ;)
Everything alright?

MatthieuDartiailh:
ping @hgrecco. fixes for #83

MatthieuDartiailh:
@joshsleeper could try to fix the merge conflicts ? Pyvisa is turning to an organization which will hopefully make the workflow smoother. Being now a member I will try to push your fix in as soon as possible (and as soon as you can fix the merge conflict). @hgrecco will member like me will be able to trigger bors once you set it in place ?

hgrecco:
@MatthieuDartiailh Should work but let's give it a try. I would suggest
1) Fork the repo,
2) add yourself to the AUTHORS and issue a commit
3) Make a PR against the pyvisa organization repo
4) Comment in the PR with the following text
bors r+

MatthieuDartiailh:
@hgrecco could you close now that #85 (which is just a duplicate) is merged. (Also could we have the rights to close issues and PR ?)

hgrecco:
You should have the rights to do so! Let me check again

hgrecco:
Can you check now?

MatthieuDartiailh:
Now I can. Thanks

PR #86, opened 2016-07-08
Fixed a bug where encode() was being called on a byte array

marekp0:
`reader()` returns a byte array, so calling `reader().encode('ascii')` resulted in an error, which was fixed by simply removing the call to `encode()`.


hgrecco:
I think you are right but we might need to look around other "plugins" using this, as well as Python 2 and 3 compatibility.

skrchnavy:
This is duplicate of #66. One of these PRs shall be accepted.
`encode()` call shall be part of `reader` argument (function type).

MatthieuDartiailh:
Closed by #119

PR #100, opened 2017-06-21
TCPIP error handling fixes: bis of #85

MatthieuDartiailh:
This is #85 in which I simply took care of the merge conflicts ("" strings vs '' strings). This had already been reviewed several months ago and accepted. If nobody triggers this week-end I will do it on monday.

Added error handling for VisaIOErrors, raised with the appropriate error code.
Added error handling for socket.timeout, raised when the TCPIP-connected instrument stops responding. This could be due to a network issue or app crash, but it covers the case where the socket is left open but no communication is happening.
Minor code cleanup, removing unnecessary .split() functions for example.


MatthieuDartiailh:
Merging now that Travis is working again.
bors r+

Issue #61, opened 2015-10-08
Deadlock on Read with TCPIP Socket


mluthi:
In debugging a problem with reading data from a device using the TCPIP socket connection I discovered a problem in the read method of the TCPIPSocketSession class (ln. 391ff in tcpip.py; version 0.3.dev0): the select function is called without a timeout and therefore may block indefinitely in case the VI_ATTR_TERMCHAR_EN attribute is not set. In my case the while loop is executed once in the beginning to receive the data and then blocks forever the second time it gets to the select call.

My environment: Windows 7 Enterprise, SP1, Python 3.4.1

Best Regards,
Matthias

petervago:
I experienced the same. (Ubuntu 14.04. 64bit)
The following solved the issue for me:

Replaced the following code:
/usr/local/lib/python3.4/dist-packages/pyvisa-py/tcpip.py line 391

`select.select([self.interface], [] , [])`

with this:

`try:
 select.select([self.interface], [] , [] , 0.01) # timeout 0.01sec
 last = read_fun(chunk_length)
except:
 last = ""`

skrchnavy:
in #107 there shall be implemented fix for this issue. Tested on windows only for now.

skrchnavy:
#107 merged, #113 improves implementation. When merged or closed, this issue can be also closed as resolved.

MatthieuDartiailh:
Closed by #113

PR #107, opened 2017-10-26
TCPIP handler read shall not block

skrchnavy:
tcpip handler was blocking in select.

this implementation waits for data for `timeout` seconds when in loop for first time, then _polls_ in 0.01 sec interval. (sleep removed).

This shall solve #61

### DIFF
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -385,10 +385,7 @@ class TCPIPSocketSession(Session):
         select_timout = min(timeout, 2.0)
         while now - start <= timeout:
             # use select to wait for read ready, max `select_timout` seconds
-            try:
-                r, w, x = select.select([self.interface], [], [], select_timout)
-            except KeyboardInterrupt as err:
-                raise err
+            r, w, x = select.select([self.interface], [], [], select_timout)
 
             last = b''
             if self.interface in r:
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -381,8 +381,8 @@ class TCPIPSocketSession(Session):
             return (out + parts[0] + end_byte,
                     constants.StatusCode.success_termination_character_read)
 
-        # initial 'select_timout' is same as timeout, so when no data arrived then max block time
-        select_timout = timeout
+        # initial 'select_timout' is same as timeout or max 2s, so when no data arrived then max block time
+        select_timout = min(timeout, 2.0)
         while now - start <= timeout:
             # use select to wait for read ready, max `select_timout` seconds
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Looks good to me. Is this ready to merge ?

skrchnavy:
Need to test on linux and I am not sure if KeyboardInterrupt is handled properly (if it shall not rethrow the exception.

skrchnavy:
KeyboardInterrupt is reraised. Also timeout is calculated as float to catch miliseconds properly (1900 ms was handled as 1 sec not 1.9 sec).

ready for final review / merge

pyvisa-py/tcpip.py
```diff
@@ -365,7 +365,7 @@ def read(self, count):
         end_char, _ = self.get_attribute(constants.VI_ATTR_TERMCHAR)
         enabled, _ = self.get_attribute(constants.VI_ATTR_TERMCHAR_EN)
         timeout, _ = self.get_attribute(constants.VI_ATTR_TMO_VALUE)
-        timeout /= 1000
+        timeout = float(timeout)/1000.0
```
MatthieuDartiailh: You could consider adding 
```python
from __future__ import division
```
to the beginning of the file which would make this useless Alternatively dividing by 1000.0 is enough no need to additionally cast to float explicitely.

pyvisa-py/tcpip.py
```diff
@@ -381,13 +381,23 @@ def read(self, count):
...
+        # initial 'select_timout' is same as timeout or max 2s, so when no data arrived then max block time
+        select_timout = min(timeout, 2.0)
         while now - start <= timeout:
-            # use select to wait for read ready
-            select.select([self.interface], [], [])
-            last = read_fun(chunk_length)
+            # use select to wait for read ready, max `select_timout` seconds
+            try:
+                r, w, x = select.select([self.interface], [], [], select_timout)
+            except KeyboardInterrupt as err:
```
MatthieuDartiailh: What is the point of catching the exception to re-raise it directly here ?

MatthieuDartiailh:
Minor things I would like to see fixed before merging. Please note that I will be away next week.

pyvisa-py/tcpip.py
skrchnavy: addressed

pyvisa-py/tcpip.py
skrchnavy: On windows select is no unblocked with KeyboardInterrupt but waits for timeout to be expired. But I tested it without catching the exception and it works OK, so removed try/except block.

MatthieuDartiailh:
Looks good ! Triggering bors

bors r+

PR #111, opened 2017-11-23
pyvisa-py moved to pyvisa organization

skrchnavy:
organization was changed from hgrecco to pyvisa

related: pyvisa/pyvisa#265 and pyvisa/pyvisa#294


MatthieuDartiailh:
Thanks @skrchnavy.

bors r+

PR #113, opened 2017-11-24
TCPIP timeout handling improvement

skrchnavy:
Improvement of #107

Code is simplified in calculation when timeout elapsed. (removed variable now)

### _poll_ interval calculation is improved.
_Note: Poll interval is necessary on windows, because select is not 'broken' by signal. so waiting in select for example 10 seconds would block program for 10 seconds until signal (CTRL+C) processed. \
On linux there is no such issue with ._

poll_ interval is calculated as 50% of previous _poll_ interval (min is 50ms). \
Previous constant 10ms _poll_ interval could increase CPU utilization.

When merged, #61 can be closed

### DIFF
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -371,7 +371,6 @@ class TCPIPSocketSession(Session):
 
         read_fun = self.interface.recv
 
-        now = start = time.time()
 
         out = self._pending_buffer
 
@@ -383,7 +382,9 @@ class TCPIPSocketSession(Session):
 
         # initial 'select_timout' is same as timeout or max 2s, so when no data arrived then max block time
         select_timout = min(timeout, 2.0)
-        while now - start <= timeout:
+        # time, when loop shall finish
+        finish_time = time.time() + timeout
+        while time.time() <= finish_time:
             # use select to wait for read ready, max `select_timout` seconds
             r, w, x = select.select([self.interface], [], [], select_timout)
 
@@ -395,7 +396,6 @@ class TCPIPSocketSession(Session):
                 # can't read chunk or timeout
                 # `select_timout` decreased to 0.01 sec
                 select_timout = 0.01
-                now = time.time()
                 continue
 
             if enabled and end_byte in last:
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -394,8 +394,8 @@ class TCPIPSocketSession(Session):
 
             if not last:
                 # can't read chunk or timeout
-                # `select_timout` decreased to 0.01 sec
-                select_timout = 0.01
+                # `select_timout` decreased to 20% of previous, min wait time is 100ms
+                select_timout = max(select_timout/5.0, 0.1)
                 continue
 
             if enabled and end_byte in last:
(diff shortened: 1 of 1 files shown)


pyvisa-py/tcpip.py
```diff
@@ -381,21 +380,22 @@ def read(self, count):
...
             last = b''
             if self.interface in r:
                 last = read_fun(chunk_length)
 
             if not last:
                 # can't read chunk or timeout
-                # `select_timout` decreased to 0.01 sec
-                select_timout = 0.01
-                now = time.time()
+                # `select_timout` decreased to 50% of previous, min wait time is 100ms
```
MatthieuDartiailh: min is 50 ms, no ?

skrchnavy: yes, I tested 100ms and 50ms for 'timeout delay', so it is timeouted in interval '<timeout, timeout + 50ms)', typicaly <40ms delay.

MatthieuDartiailh:
LGTM. Can I assume you checked this works as expected ?

skrchnavy:
yes, i tested it on windows 7 and works as described. 

With 10s timeout, it waited 2secs then ~1 sec, then ~0.5 sec, etc. and finished after 123 iterations, table shows time missing to timeout (measured after select unblocked)
```
8.0
6.99799990654
6.4960000515
6.24300003052
6.11599993706
6.05099987984
5.99900007248
5.94799995422
5.89700007439
==== CUT ====
0.148000001907
0.095999956131
0.0439999103546
-0.00799989700317
```
Last value shows that function finshed 8ms later then planed.

With 3s timeout, measurement is:
```
1.5
0.748000144958
0.371000051498
0.180999994278
0.085000038147
0.0329999923706
-0.0190000534058
```
finished 19ms after planed timeout.

MatthieuDartiailh:
Thinking a bit more about this, I would like to see a small modification to provide a better support of short timeout (sometimes useful in debugging). Could you add a `min_select_timeout = max(min(select_timeout/10, 50), 1)` , and use it instead of 50, such that we get something closer to the expected timeout for timeout in the range 1 to 50 ms.

To fix the conflicts please rebase on master rather than merging it makes for a cleaner history.

MatthieuDartiailh:
There are still some conflicts that needs to be solved, once it is done I will merge.

skrchnavy:
I was not able locally rebase changes so I did the merge. 
I will do final change in calculation of select_timeout and min_select_timeout and then you can review / merge

skrchnavy:
ready for final review.
measurments  (diff is difference between real timeout and expected timeout, it shall be <1% from expeted timeout):
* timeout=1s
```select_timout=0.5s
select_timout=0.25s
select_timout=0.125s
select_timout=0.0625s
select_timout=0.03125s
select_timout=0.015625s
select_timout=0.01s
diff=7.999897003173828ms
```
* timeout=2s
```select_timout=1.0s
select_timout=0.5s
select_timout=0.25s
select_timout=0.125s
select_timout=0.0625s
select_timout=0.03125s
select_timout=0.02s
diff=3.000020980834961ms
```
* timeout 10s
```select_timout=2.0s
select_timout=1.0s
select_timout=0.5s
select_timout=0.25s
select_timout=0.125s
select_timout=0.1s
select_timout=0.1s
select_timout=0.1s
        ...
select_timout=0.1
diff=95.99995613098145ms
```

MatthieuDartiailh:
I just added a tiny bit more docs (in case somebody in the future think about reverting to only use select timeout). If you are happy with the change, you can trigger bors

bors delegate+

skrchnavy:
Agreed and approved

Issue #109, opened 2017-11-05
Unsupported attributes for PyVISA-py TCPIP SOCKET communication


MatthieuDartiailh:
First reported on PyVisa: https://github.com/pyvisa/pyvisa/issues/274

skrchnavy:
@cwgamma 
I checked documentation for NI: http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/vi_attr_send_end_en/
`VI_ATTR_SEND_END_EN` is not valid for TCPIP SOCKET only for TCPIP INSTR.

Anyway, it works as you have written, I tested it using pyvisa shell with backend pyvisa-py, attribute is supported but not listed and can't be changed.

MatthieuDartiailh:
Should we close then ?

skrchnavy:
no, this is real issue, VI_ATTR_SEND_END_EN shall be ignored but ['VI_ATTR_SUPPRESS_END_EN'](http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/vi_attr_suppress_end_en/) shall be implemented. I am working on it for TCPIP.

MatthieuDartiailh:
Cool

MatthieuDartiailh:
Closed by #114

PR #114, opened 2017-11-27
VI_ATTR_SUPPRESS_END_EN in TCPIP SOCKET

skrchnavy:
attribute is set and read in TCPIP SOCKET resource driver. It works similar to description in http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/vi_attr_suppress_end_en/ :\
_On TCP/IP SOCKET sessions, if this attribute is set to VI_FALSE, if NI-VISA reads some data and then detects a pause in the arrival of data packets, it will terminate the read operation. On TCP/IP SOCKET sessions, this attribute defaults to VI_TRUE in NI-VISA._

relates to #109


skrchnavy:
This could be in conflict with #113 so when one accepted, second shall be check for proper merge.

MatthieuDartiailh:
I am happy with. Merging.

bors r+

MatthieuDartiailh:
By the way could you request me as a reviewer when you open a PR that way I can navigate more quickly to them (and I am hence less likely to forget about them).

Issue #88, opened 2016-09-23
opening TCP/IP connection to invalid IP blocks forever


iliis:
While `resource_manager.open_resource('TCPIP::127.0.0.1')` immediately raises a `ConnectionRefusedError`, connecting to an IP that doesn't exist (e.g. doing  `resource_manager.open_resource('TCPIP::192.255.255.42')` or something) will **block forever**.

This is an issue in https://github.com/hgrecco/pyvisa-py/blob/master/pyvisa-py/protocols/rpc.py#L339: `socket.connect()` blocks as the socket is in blocking mode. Adding for example `socket.settimeout(5)` immediately before this puts the socket into timeout mode and the `connect()` will timeout after five seconds.

I haven't checked, but as long as the socket is in blocking mode other operations might behave similarly and block forever (or timeout after a very long time).

I would suggest to at least set some default timeout. Pyvisa actually has a parameter for that in `open_resource()` called `open_timeout` (see http://pyvisa.readthedocs.io/en/stable/api/resourcemanager.html#pyvisa.highlevel.ResourceManager.open_resource).

See also https://docs.python.org/3/library/socket.html#socket-timeouts

skrchnavy:
Different approach implemented in TCPIP INSTR and TPCIP SOCKET.
* TCPIP INSTR:
```
    def connect(self):
        logger.debug('RawTCPClient: connecting to socket at (%s, %s)', self.host, self.port)
        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self.sock.connect((self.host, self.port))
 ```
* TPCIP SOCKET:
```
        self.interface = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self.interface.setblocking(0)

        try:
            self.interface.connect_ex((self.parsed.host_address, int(self.parsed.port)))
        except Exception as e:
            raise Exception("could not create socket: %s" % e)
```

None is correct, handling of connect timeout is not provided. In SOCKET mode, also there is no check, if connect was succesful.

skrchnavy:
#115 implemented connect timeout in TCPIP SOCKET session.
@iliis Is same expected also in TCPIP INSTR session? (I don't have access to such device so can't test).

iliis:
@skrchnavy Uhm, I'm not sure. What's the difference? I'm sending GPIB commands to the device via TCP, so its INSTR?
[...]

Issue #101, opened 2017-07-10
Problem with TCPIP read() when more than "chunk_length" is returned


rriggs:
I am attempting to download a screenshot from a Siglent DSO using the SCDP command.  This just dumps the binary data for a BMP file.  The problem appears to be that it sends all of the data in one shot.

```python
n, status = scope.write('SCDP')
if status == StatusCode.success:
    data = scope.read_raw()
    open("test.bmp", "w").write(data)
```
The code just hangs at read_raw().

Adding some diagnostics to TCPIPInstrSession.read(), this is what I see:

```
wanted 4096 bytes, read 1152057 bytes
want -1131577 more bytes
```

Upon return from this function, count is negative, the excess data read and returned, but the count is lost in the upper layers.  I don't understand the VXI-11 protocol well enough to know if this is a problem with the device not conforming to spec (it seems to work with NI-VISA and their native app), with PyVISA or with PyVISA-py.

I bodged the read() function to buffer the data it receives and only return the requested chunk_size.  That seems to work for me (I successfully stored a screenshot), but I don't know if that is a reasonable fix.  If it is, I am happy to work up a pull request.

rriggs:
It appears I can work around the issue by telling read_raw() to read 2MB of data.  Somehow I missed that read_raw() accepts an optional length argument.

skrchnavy:
@rriggs are you able test this isse when #116 merged?

skrchnavy:
#116 merged, is there a chance to verify if this issue is fixed?

PR #115, opened 2017-11-27
TCPIP SOCKET improvements

skrchnavy:
* Connect uses open_timeout parameter to be timeouted (#88)
* Refactored read operation with improved timeout handling
* Read Uses bytearray and bytes instead of string - inspired by #50 
* Calculation of return data in read uses indexes
* Renamed some variables for better readability.

### DIFF
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -394,14 +394,14 @@ class TCPIPSocketSession(Session):
             # use select to wait for read ready, max `select_timout` seconds, min is 'min_select_timeout' seconds
             r, w, x = select.select([self.interface], [], [], select_timout)
 
-            last = b''
+            read_data = b''
             if self.interface in r:
-                last = read_fun(chunk_length)
+                read_data = read_fun(chunk_length)
 
-            if last:
-                out.extend(last)
+            if read_data:
+                out.extend(read_data)
 
-                if term_char_en and term_byte in last:
+                if term_char_en and term_byte in read_data:
                     term_byte_index = out.index(term_byte) + 1
                     self._pending_buffer = out[term_byte_index:]
                     return bytes(out[:term_byte_index]), constants.StatusCode.success_termination_character_read
--- a/pyvisa-py/tcpip.py
+++ b/pyvisa-py/tcpip.py
@@ -381,12 +381,13 @@ class TCPIPSocketSession(Session):
             self._pending_buffer = out[term_byte_index:]
             return bytes(out[:term_byte_index]), constants.StatusCode.success_termination_character_read
 
-        # On Windows, select is not interrupted by KeyboardInterrupt, to
-        # avoid blocking for very long time, we use a decreasing timeout
-        # in select
-        # minimum select timeout to avoid too short select interval (minimum is in interval 1 - 100ms based on timeout)
+        # On Windows, select is not interrupted by KeyboardInterrupt, to avoid blocking
+        # for very long time, we use a decreasing timeout in select
+        # minimum select timeout to avoid too short select interval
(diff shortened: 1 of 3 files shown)


skrchnavy:
@MatthieuDartiailh please review (I can't change reviewers).

MatthieuDartiailh:
By the way mentioning me in the PR also makes it easy to find, so you can keep doing this while waiting for hgrecco to add you.

skrchnavy:
The part at line 379 is something hat shall be considered. This is good for handling chunks or binary data when buffer not big enough.
When termchar presented, in normal situation I don't expect further data shall be processed (using standard query). Special case could be in case of using send and read instead of query, when called 2 times send and then 2 times query.

skrchnavy:
this could be a fix of #101

MatthieuDartiailh:
Nice refactoring !

skrchnavy:
@MatthieuDartiailh  I have added handling of timeout in connect (#88)

skrchnavy:
There are few improvements possible in processing return codes from connect_ex but it is difficult to identify errno codes for all systems, so for now I see implementation sufficient. Select waits 100ms min and it is usually enough to connect to a resource on LAN.

Merging
bors r+

MatthieuDartiailh:
Just one thing I missed and that need to be fixed: the `_pending_buffer` defined at https://github.com/pyvisa/pyvisa-py/blob/master/pyvisa-py/tcpip.py#L327 should not be a class variable since otherwise it will be shared by all instances, which not what we want at all ! this attribute should be created in `__init__`

skrchnavy:
I was thinking about it, and not only pending_buffer but also other variables. Fortunately not used in code.
By mu opinion, intention was not to implement `__init__` in all sub-classes and call `super` and have variables.

MatthieuDartiailh:
There is a number of class variables that are indeed perfectly fine because they are simply used to alter some behaviors easily on subclasses. However pending_buffer HAS to be an instance variable as otherwise it would be shared by alll instances of TCPIPSocket which is not desirable ! So either we initialize it in init or on first use (but i do not favor this solution).

skrchnavy:
I agree, I am cleaning such variables in other features when I am able to test them.
For pending buffer I have to find such device (function) that is able to produce more data then are consumed in read  or simulate such device (pending buffer is in action).

skrchnavy:
@MatthieuDartiailh please check #116 for implementation of pending buffer per session.

pyvisa-py/tcpip.py
```diff
@@ -364,64 +398,59 @@ def read(self, count):
...
+
+            # check, if we have any data received (from pending buffer or further reading)
+            if term_char_en and term_byte in out:
+                term_byte_index = out.index(term_byte) + 1
+                self._pending_buffer = out[term_byte_index:]
+                return bytes(out[:term_byte_index]), constants.StatusCode.success_termination_character_read
+
+            if len(out) >= count:
+                self._pending_buffer = out[count:]
+                return bytes(out[:count]), constants.StatusCode.success_max_count_read
```
skrchnavy: this shall be a fix of #101 , returning only expected bytes

MathieuLeocmach:
Thanks a lot for these feature improvements. 

With the version one can get with pip (0.2) I was experiencing infinite waiting when reading from a TCPIP socket that transmit binary data without termination character. I had identified a timeout problem in `pyvisa-py.tcpip.TCPIPSocketSession.read`, but I was unable to go further. Taking the bleeding edge from github solved the problem.

MatthieuDartiailh:
Happy to know this works for you. Hopefully we will be able to make a new release soon.

PR #117, opened 2017-11-30
Supported python 3.5 and 3.6

skrchnavy:
added required versions into setup.py to unify with pyvisa.


skrchnavy:
@MatthieuDartiailh ready for final review

Issue #73, opened 2015-12-15
AttributeError: 'bytes' object has no attribute 'encode' when reading from GPIB


iliis:
Executing this code:

``` python
import visa
visa.log_to_screen()
rm = visa.ResourceManager('@py')
spectrum_analyzer = rm.open_resource('GPIB0::20::INSTR')
spectrum_analyzer.write("*IDN?")
print(spectrum_analyzer.read_raw())
```

Fails with:

```
2015-12-15 16:01:40,178 - pyvisa - DEBUG - SerialSession was correctly imported.
2015-12-15 16:01:40,203 - pyvisa - DEBUG - USBSession and USBRawSession were correctly imported.
2015-12-15 16:01:40,209 - pyvisa - DEBUG - TCPIPSession was correctly imported.
2015-12-15 16:01:40,212 - pyvisa - DEBUG - GPIBSession was correctly imported.
2015-12-15 16:01:40,212 - pyvisa - DEBUG - Created library wrapper for unset
2015-12-15 16:01:40,213 - pyvisa - DEBUG - Created ResourceManager with session 4163272
2015-12-15 16:01:40,213 - pyvisa - DEBUG - GPIB0::20::INSTR - opening ...
2015-12-15 16:01:40,217 - pyvisa - DEBUG - GPIB0::20::INSTR - is open with session 9839346
2015-12-15 16:01:40,217 - pyvisa - DEBUG - GPIB.write b'*IDN?\r\n'
2015-12-15 16:01:40,226 - pyvisa - DEBUG - GPIB0::20::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
Traceback (most recent call last):
  File "./measure.py", line 21, in <module>
    print(spectrum_analyzer.read_raw())
  File "/usr/lib/python3.4/site-packages/pyvisa/resources/messagebased.py", line 306, in read_raw
    chunk, status = self.visalib.read(self.session, size)
  File "/usr/lib/python3.4/site-packages/PyVISA_py-0.3.dev0-py3.4.egg/pyvisa-py/highlevel.py", line 253, in read
    return self.sessions[session].read(count)
  File "/usr/lib/python3.4/site-packages/PyVISA_py-0.3.dev0-py3.4.egg/pyvisa-py/gpib.py", line 128, in read
    return self._read(reader, count, checker, False, None, False, gpib.GpibError)
[...]

Issue #69, opened 2015-11-03
UnicodeDecodeError with USB INSTR but not with TCPIP INSTR


in3otd:
Hello,
when trying to read the trace binary data from a Rigol scope using USB I get often errors like `UnicodeDecodeError: 'ascii' codec can't decode byte 0x80 in position 12: ordinal not in range(128)`. Sometimes I get this error even when the instrument should have sent a real ASCII string, but it might be actually sending invalid bytes due to scope FW bugs. Anyway, I think the `UnicodeDecodeError` should not happen in any case.
When I connect using Ethernet this error does not happen.

With the log enabled this is what I get:

```
2015-11-03 19:35:22,312 - pyvisa - DEBUG - SerialSession was correctly imported.
2015-11-03 19:35:22,342 - pyvisa - DEBUG - USBSession and USBRawSession were correctly imported.
2015-11-03 19:35:22,344 - pyvisa - DEBUG - TCPIPSession was correctly imported.
2015-11-03 19:35:22,345 - pyvisa - DEBUG - GPIBSession was correctly imported.
2015-11-03 19:35:22,345 - pyvisa - DEBUG - Created library wrapper for unset
2015-11-03 19:35:22,345 - pyvisa - DEBUG - Created ResourceManager with session 6034227
2015-11-03 19:35:22,345 - pyvisa - DEBUG - USB0::6833::1230::DS1ZC162943448::0::INSTR - opening ...
2015-11-03 19:35:22,508 - pyvisa - DEBUG - USB0::6833::1230::DS1ZC162943448::0::INSTR - is open with session 9064266
24000000
2015-11-03 19:35:24,013 - pyvisa - DEBUG - USB0::6833::1230::DS1ZC162943448::0::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
act_md= u'24000000\n\x7f' md= 24000000
Memory depth set to 24000000
0
2015-11-03 19:35:25,016 - pyvisa - DEBUG - USB0::6833::1230::DS1ZC162943448::0::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
Traceback (most recent call last):
  File "mytest.py", line 10, in <module>
    cd = s.read()
  File "/home/username/mylib.py", line 67, in read
    d = self.i.read_raw()
[...]

PR #119, opened 2017-12-01
Fix of encode called on bytes object

skrchnavy:
this is PR to fix PR #59 in gpib module and not in generic session when reader shall return bytes and not string. similar approach was provided in #70 (author @DavidFabijan )
It shall fix these issues: #69, #73,  

Problem in gpib  could be not solved perfectly, but is not spread across other session types.

This invalidates PRs: #66 (author @herm ) , #70 (@DavidFabijan ), #86 (@marekp0). All could be closed.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -125,7 +125,7 @@ class GPIBSession(Session):
         # 0x2000 = 8192 = END
         checker = lambda current: self.interface.ibsta() & 8192
 
-        reader = lambda: self.interface.read(count)
+        reader = lambda: self.interface.read(count).encode('ascii')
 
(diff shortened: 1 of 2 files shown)


skrchnavy:
@MatthieuDartiailh _read operation is candidate for small refactoring, I have idea to use it also in TCPIP SOCKET session. So for now we shall do this small step.

MatthieuDartiailh:
This looks good to me. Should I merge now ?

skrchnavy:
If not further comments, please merge

MatthieuDartiailh:
Merging. Closes #70, #66, #86, #59, #69, #73 

bors r+

PR #121, opened 2017-12-14
Update highlevel.py

elkevn:
Modified the list_ressources() function, which was not coherent with the pyvisa.ctwrapper file.
Instead of sending back an error if no ressource is found, there should be an empty tuple sent back.


MatthieuDartiailh:
Looks good to me. Thanks for taking the time to contribute.

bors r+

PR #120, opened 2017-12-06
TCPIP INSTR Session / RPC timeout handling improvements

skrchnavy:
* `rpc` module changes:
  * updated to use timeout and fragmentation size in recv and write operations
  * fragmentation support in write operation
  * connect is timeouted based on timeout
  * 'select' moved to `_sendrecord` and `_recvrecord` functions
* `tcpip` module:
 * open_timeout passed to rpc via vxi11

Related to #88

PR #126, opened 2018-02-09
Clean up and fix a nasty bug in tcpip instr session.

MatthieuDartiailh:
The term char was incorrectly converted, which lead to read incomplete messages.
Also optimize some routines to use bytearray and some general pep 8 formatting, and prevent a hard fail if one instrument connected in USB is not properly configured (can happen on windows with the wrong driver installed)

@skrchnavy if you have time to review it would be nice. This fixes the issue I have seen in #120 .

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -86,8 +86,9 @@ Now:
 Are all VISA attributes and methods implemented?
 ------------------------------------------------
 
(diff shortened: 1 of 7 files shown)


MatthieuDartiailh:
Merging this is really an improvement.

bors r+

PR #130, opened 2018-03-28
Attempt to fix an issue in rpc in which a select timeout cause a EOFError

MatthieuDartiailh:
The basic idea is that compared to the old version (pre PR 67), we can now get an empty read_data because the select timed out, which by itself may no mean anything, and we need to wait to see if we get a real timeout rather than complaining about EOFError early.

@svanan77 could you review and test. Hopefully this should fix https://github.com/pyvisa/pyvisa/issues/319 and given the nature of the issue, it is possible I could not easily reproduce locally.

### DIFF
--- a/pyvisa-py/protocols/rpc.py
+++ b/pyvisa-py/protocols/rpc.py
@@ -353,6 +353,9 @@ def _recvrecord(sock, timeout, read_fun=None):
             buffer.extend(read_data)
         elif timeout is not None and time.time() >= finish_time:
             # reached timeout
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
I removed a small of the debug that I fear may be a bit costly to keep. This is ready to merge as far as I am concerned since this is more correct than the previous version even if it does not appear to fix https://github.com/pyvisa/pyvisa/issues/319 in all cases. Any review welcome, and otherwise I will merge sometimes next week.

PR #164, opened 2018-10-09
Use the common logger for USB logs

henriknil:
Using the common logger makes it possible to redirect the logs to stdout on
demand instead of by default.


MatthieuDartiailh:
Thanks @henriknil and sorry for the delay. I missed this one for some time. Do not hesitate to ping me if this happens again in the future.

PR #166, opened 2018-11-20
USB: fix error with infinite timeout in interface

mgubser:
If on ``self.interface`` the timeout is never set or explicitly set to ``None`` this function will raise an exception on line 77 (``TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'``). Therefore I suggest this change, which was probably the intent of the author.

### DIFF
--- a/pyvisa-py/usb.py
+++ b/pyvisa-py/usb.py
@@ -71,7 +71,7 @@ class USBSession(Session):
 
     def _get_timeout(self, attribute):
         if self.interface:
-            if self.interface.timeout == 2**32-1:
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Thanks for catching this. However it turns out the proposed fix is not the correct one. When the timeout of the USB interface is set to `None` it will use the default timeout of pyusb which is not what we want. The proper fix is to make sure we pass the expected timeout to pyusb when creating the interface, which was not what was happening. I implemented the proper fix in a new pull request. However since I do not have the hardware required to test currently I would appreciate if you could test it.

PR #167, opened 2018-11-20
sessions: pass the timeout value when creating a USB interface.

MatthieuDartiailh:
Proper fix for #166 

@mgubser could you please review and test.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,9 @@ PyVISA-py Changelog
 0.4.0 (unreleased)
 ------------------
 
-
+- fix initialization of timeout for the USB resources (the default was set
+  before creating the underlying connection to which the timeout must be passed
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
@mgubser thanks for catching this actually I got confused (once again) by the purpose of timeout on session. Hopefully this time I made the explanation clear enough (and got it right). The interface timeout should always be an integer and in ms which is why I reassign the default value after creating the interface rather than simply converting to int as you suggested in #168 . Please check this fixes your issue.

mgubser:
@MatthieuDartiailh This looks ok now. I don't see any related errors anymore. Thanks for taking on this issue.

MatthieuDartiailh:
Thanks @mgubser. I will update the release notes and merge sometime this week.

Issue #169, opened 2018-11-26
assert_trigger() is defined twice


tivek:
https://github.com/pyvisa/pyvisa-py/blob/5f41742ca41beb0701d879558d2a5db5924a8485/pyvisa-py/highlevel.py#L246

https://github.com/pyvisa/pyvisa-py/blob/5f41742ca41beb0701d879558d2a5db5924a8485/pyvisa-py/highlevel.py#L198

assert_trigger() is doubly implemented by #136 and #139

MatthieuDartiailh:
Good catch ! Feel free to open a PR to fix it if you have the bandwidth.

tivek:
Sure, I can do that. I have a related question that I would like to go through with you first. When an invalid session is used, some PyVisaLibrary functions return the tuple `0, constants.StatusCode.error_invalid_object`, while others return just `constants.StatusCode.error_invalid_object`. The latter seems to be more in line with what the NI backend does. I am inclined to return just a simple return code. What do you think? If you agree, the remaining functions that return tuples should also be touched up.

MatthieuDartiailh:
At the C level in VISA, all functions return only the status. But pointers are used to pass additional information.

On the Python side, we usually return the status code and whatever value would have been returned through a pointer (the status code is usually returned last). You can have a look at the implementation of write for example (https://github.com/pyvisa/pyvisa/blob/master/pyvisa/ctwrapper/functions.py#L1852) and generally the proper signature for the library methods are defined in https://github.com/pyvisa/pyvisa/blob/master/pyvisa/highlevel.py#L374

I hope this clarifies the situation.

tivek:
Thanks @MatthieuDartiailh. The question was originally about `assert_trigger()` since the two copies had different return types. But, as far as I can tell there is more. The following return type signatures are different than specified in `VisaLibraryBase` (pyvisa master):
- `PyVisaLibrary.assert_trigger()` - should return `StatusCode`, so it seems the second of two copies is correct,
- `PyVisaLibrary.clear()` - should return `StatusCode`, returns `0, StatusCode`,
- `PyVisaLibrary.gpib_send_ifc()` - should return `StatusCode`, returns `0, StatusCode`,
- `Session.gpib_command()` - should return `int, StatusCode`, returns `StatusCode`,
- `GPIBSession.clear()` - should return `StatusCode`, returns `0, StatusCode`,
[...]

PR #170, opened 2018-11-27
fix return types

tivek:
Fix return types to be along the lines of pyvisa.highlevel.VisaLibraryBase, as discussed in #169, plus some small corrections.

### DIFF
--- a/pyvisa-py/sessions.py
+++ b/pyvisa-py/sessions.py
@@ -262,14 +262,10 @@ class Session(compat.with_metaclass(abc.ABCMeta)):
 
         :param command_byte: command byte to send
         :type command_byte: int, must be [0 255]
-        :return: return value of the library call
-        :rtype: :class:`pyvisa.constants.StatusCode`
+        :return: Number of written bytes, return value of the library call.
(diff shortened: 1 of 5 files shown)


pyvisa-py/sessions.py
```diff
@@ -262,14 +262,10 @@ def gpib_command(self, command_byte):
...
-        :rtype: :class:`pyvisa.constants.StatusCode`
+        :return: Number of written bytes, return value of the library call.
+        :rtype: int, :class:`pyvisa.constants.StatusCode`
         """
-        try:
-            return self.sessions[session].gpib_command(command_byte)
-
-        except KeyError:
-            return constants.StatusCode.error_invalid_object
+        return StatusCode.error_nonsupported_operation
```
MatthieuDartiailh: This looks incorrect to me we need to return `0, StatusCode.error_nonsupported_operation` as per the docstring which is correct.

tivek: Correct, I missed that one. Will update the PR.

MatthieuDartiailh:
Merging

bors r+

PR #171, opened 2018-11-27
fix double close of GPIB interface and controller in GpibSession

tivek:
Fix a double close of GPIB connections which usually produces uncaught exceptions.

To reproduce the bug on pyvisa-py master:
```python
$ ipython
Python 3.7.0 (default, Oct  9 2018, 10:31:47) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.1.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import visa                                                                                                                                                                                                                           

In [2]: rm = visa.ResourceManager('@py')                                                                                                                                                                                                      

In [3]: tek = rm.open_resource('GPIB0::30::INSTR')                                                                                                                                                                                            

In [4]: tek.query('*IDN?')                                                                                                                                                                                                                    
Out[4]: 'TEKTRONIX,TDS 3014C,0,CF:91.1CT FV:v4.05 TDS3GV:v1.00 TDS3FFT:v1.00 TDS3TRG:v1.00\n'

In [5]: quit()                                                                                                                                                                                                                                
libgpib: invalid descriptor
libgpib: invalid descriptor
Exception ignored in: <function Gpib.__del__ at 0x7f985e1996a8>
  File "/home/tivek/projects/linux-gpib-code/linux-gpib-user/language/python/Gpib.py", line 32, in __del__
gpib.GpibError: close() failed: One or more arguments to the function call were invalid.
```

Explanation and fix:
GpibSession uses two Gpib.Gpib objects (linux-gpib project) to manage connections to the instrument and GPIB board. These clean up on their own in `__del__()` and we should not manually close their handles.

Long-term, `__del__()` should not be used for reliable finalization since it is a potentially ugly implementation detail of CPython's GC. Sadly, Gpib.Gpib does not expose a better interface for cleanup. At some point we should either patch/fork Gpib.Gpib or switch to classic gpib function calls with integer handles.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -20,6 +20,17 @@ from .sessions import Session, UnknownAttribute
 try:
     import gpib
     from Gpib import Gpib
+
+    # patch Gpib to avoid double closing of handles
+    def _patch_Gpib():
+        if not hasattr(Gpib, "close"):
+            _old_del = Gpib.__del__
+            def _inner(self):
+                _old_del(self)
+                self._own = False
+            Gpib.__del__ = _inner
+            Gpib.close = _inner
+    _patch_Gpib()
 except ImportError as e:
     Session.register_unavailable(constants.InterfaceType.gpib, 'INSTR',
                                  'Please install linux-gpib to use this resource type.\n%s' % e)
@@ -125,10 +136,8 @@ class GPIBSession(Session):
         return status
 
     def close(self):
-        del self.interface
-        del self.controller
-        self.interface = None
-        self.controller = None
+        self.interface.close()
+        self.controller.close()
 
     def read(self, count):
         """Reads data from device or interface synchronously.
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Indeed this looks like a bad idea. At least the __del__ should be safe if the interface is closed. Could you implement the fix in gpib_ctypes and we could then switch to this implementation, since we do not win anything in using linux-gpib binding ?

MatthieuDartiailh:
I will merge sometimes this week since this is no matter what the sensible solution for now.

tivek:
linux-gpib has [merged a patch](https://sourceforge.net/p/linux-gpib/code/1768/) that introduces Gpib.close() like gpib_ctypes 0.2.0. Until it gets released, and to support users of older linux-gpib releases, I propose to keep the monkeypatching code from 5968788.

MatthieuDartiailh:
Sounds good. I agree that keeping the monkeypatch for now makes sense. Is this ready for merging ?

tivek:
Ready for merging. Tested with linux-gpib r1767 (before the accepted patch), linux-gpib trunk as well as gpib_ctypes 0.2.0.

MatthieuDartiailh:
I forgot to ask and I am going to fix it right away but in the future please update the release notes.

tivek:
Sorry about the release notes, will do next time.

PR #174, opened 2018-12-05
implement VI_ATTR_GPIB_REN_STATE for GPIBSession

tivek:
Python bindings to linux-gpib [now interface with iblines](https://sourceforge.net/p/linux-gpib/code/1772/). This allows us to query the state of REN line.


MatthieuDartiailh:
Looks good and I believe you know the meaning of those constants. Merging

bors r+

Issue #142, opened 2018-04-21
GPIBInstrument.wait_for_srq() raises NotImplementedError


tivek:
``pyvisa.resources.gpib.GPIBInstrument.wait_for_srq()`` calls ``enable_event()``, ``wait_on_event()``, and ``discard_events()`` of the underlying visalib. However, ``PyVisaLibrary`` does not implement those methods so ``NotImplementedError`` is raised.

One way to go about this is to delegate the missing methods to underlying pyvisa-py session objects and implement event support there. 

Another possibility is to notice that ``wait_for_srq()`` is quite universal - in particular, VISA Service Request events should be supported by GPIB INSTR, GPIB INTFC, GPIB-VXI INSTR, TCPIP INSTR, USB INSTR, and VXI INSTR. So, ``wait_for_srq()`` could be moved to ``pyvisa.highlevel.VisaLibraryBase``. This would make it easier to implement an overload in ``PyVisaLibrary`` before full-blown event support in pyvisa-py is ready.

MatthieuDartiailh:
I am not familiar with events in VISA I must say so I have no idea how difficult it would be to provide support for them. However as I suspect this will be quite dependent on the session (I assume it will not imply the same low level manipulations) so we can probably just add the required indirection in PyVisaLibrary and implement them on a per session basis. If you are interested on working on this you are welcome.

tivek:
Also not too familiar with VISA events but I will give it a try. I hope to get back to you soon with some basic GPIB event queue functionality, timing depends on how often I manage to get access to instruments.

eszesandras:
@tivek  the enable_event is correspond to viEnableEvent which is a function in the NI-VISA library, and the pyvisa provides wrapping to NI_VISA library but that is only available on "@ni" backend. 
Otherwise you can find these functions in "pyvisa/highlevel.py"

tivek:
After a long while without regular access to a GPIB board and instruments, I started working on per-session event support [a separate branch](https://github.com/tivek/pyvisa-py/tree/GPIB-events). Will post updates here.

MatthieuDartiailh:
Sounds great @tivek ! Please keep us posted.

tivek:
[...]

PR #175, opened 2018-12-05
[WIP] Gpib events

tivek:
Request for comments, related to Issue #142

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -493,13 +493,13 @@ class GPIBSession(Session):
 
         # if the event queue is empty, wait for more events
         if not self.event_queue:
-            old_timeout = self._get_timeout(None)
+            old_timeout, _ = self._get_timeout(None)
             self._set_timeout(None, timeout)
 
             event_mask = 0
 
             if in_event_type in (constants.VI_EVENT_IO_COMPLETION, constants.VI_ALL_ENABLED_EVENTS):
-                event_mask |= gpib.CMPL
+                event_mask |= 0x100  # CMPL
 
             if in_event_type in (constants.VI_EVENT_SERVICE_REQ, constants.VI_ALL_ENABLED_EVENTS):
                 event_mask |= gpib.RQS
@@ -513,7 +513,7 @@ class GPIBSession(Session):
             self._set_timeout(None, old_timeout)
 
             # TODO: set event attributes
-            if gpib.CMPL & event_mask & sta:
+            if 0x100 & event_mask & sta:
                 self.event_queue.append((constants.VI_EVENT_IO_COMPLETION, {}))
 
             if gpib.RQS & event_mask & sta:
@@ -523,4 +523,4 @@ class GPIBSession(Session):
             out_event_type, event_data = self.event_queue.pop()
             return out_event_type, event_data, StatusCode.error_timeout
         except IndexError:
-            return None, None, StatusCode.error_timeout
+            return in_event_type, None, StatusCode.error_timeout
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -25,6 +25,7 @@ try:
     def _patch_Gpib():
         if not hasattr(Gpib, "close"):
             _old_del = Gpib.__del__
+
             def _inner(self):
                 _old_del(self)
                 self._own = False
@@ -57,6 +58,8 @@ TIMETABLE = (0, 10e-6, 30e-6, 100e-6, 300e-6, 1e-3, 3e-3, 10e-3, 30e-3, 100e-3, 
 # TODO: Check board indices other than 0.
 BOARD = 0
 # TODO: Check secondary addresses.
+
+
 @Session.register(constants.InterfaceType.gpib, 'INSTR')
 class GPIBSession(Session):
     """A GPIB Session that uses linux-gpib to do the low level communication.
@@ -82,10 +85,20 @@ class GPIBSession(Session):
         timeout = 13
         send_eoi = 1
         eos_mode = 0
-        self.interface = Gpib(name=minor, pad=pad, sad=sad, timeout=timeout, send_eoi=send_eoi, eos_mode=eos_mode)
-        self.controller = Gpib(name=minor) # this is the bus controller device
+        self.interface = Gpib(name=minor, pad=pad, sad=sad,
+                              timeout=timeout, send_eoi=send_eoi, eos_mode=eos_mode)
+        self.controller = Gpib(name=minor)  # this is the bus controller device
         # force timeout setting to interface
-        self.set_attribute(constants.VI_ATTR_TMO_VALUE, attributes.AttributesByID[constants.VI_ATTR_TMO_VALUE].default)
+        self.set_attribute(constants.VI_ATTR_TMO_VALUE,
+                           attributes.AttributesByID[constants.VI_ATTR_TMO_VALUE].default)
+
+        # prepare set of allowed events
+        self.valid_event_types = [constants.VI_EVENT_IO_COMPLETION,
+                                  constants.VI_EVENT_SERVICE_REQ]
+
+        self.enabled_queue_events = set()
+
+        self.event_queue = []
 
     def _get_timeout(self, attribute):
         if self.interface:
@@ -150,9 +163,9 @@ class GPIBSession(Session):
         """
 
         # 0x2000 = 8192 = END
-        checker = lambda current: self.interface.ibsta() & 8192
+        def checker(current): return self.interface.ibsta() & 8192
 
-        reader = lambda: self.interface.read(count)
+        def reader(): return self.interface.read(count)
 
         return self._read(reader, count, checker, False, None, False, gpib.GpibError)
 
@@ -171,7 +184,7 @@ class GPIBSession(Session):
 
         try:
             self.interface.write(data)
-            count = self.interface.ibcnt() # number of bytes transmitted
+            count = self.interface.ibcnt()  # number of bytes transmitted
 
             return count, StatusCode.success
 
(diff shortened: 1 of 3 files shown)


pyvisa-py/highlevel.py
```diff
@@ -431,9 +470,146 @@ def unlock(self, session):
...
+                        (Constants.VI_QUEUE, .VI_HNDLR, .VI_SUSPEND_HNDLR, .VI_ALL_MECH)
+        :return: return value of the library call.
+        :rtype: :class:`pyvisa.constants.StatusCode`
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return StatusCode.error_invalid_object
+
+        # TODO: add support for VI_HNDLR, VI_SUSPEND_HNDLR, VI_ALL_MECH
```
MatthieuDartiailh: I feel like the session should filter itself for what kind of event it supports like you already did in GPIB.

pyvisa-py/highlevel.py
```diff
@@ -431,9 +470,146 @@ def unlock(self, session):
...
+                        (Constants.VI_QUEUE, .VI_SUSPEND_HNDLR, .VI_ALL_MECH)
+        :return: return value of the library call.
+        :rtype: :class:`pyvisa.constants.StatusCode`
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return StatusCode.error_invalid_object
+
+        #  TODO: add support for VI_HNDLR, VI_SUSPEND_HNDLR, VI_ALL_MECH
```
MatthieuDartiailh: Same as above

pyvisa-py/gpib.py
```diff
@@ -82,10 +85,20 @@ def after_parsing(self):
...
+        self.interface = Gpib(name=minor, pad=pad, sad=sad,
+                              timeout=timeout, send_eoi=send_eoi, eos_mode=eos_mode)
+        self.controller = Gpib(name=minor)  # this is the bus controller device
         # force timeout setting to interface
-        self.set_attribute(constants.VI_ATTR_TMO_VALUE, attributes.AttributesByID[constants.VI_ATTR_TMO_VALUE].default)
+        self.set_attribute(constants.VI_ATTR_TMO_VALUE,
+                           attributes.AttributesByID[constants.VI_ATTR_TMO_VALUE].default)
+
+        # prepare set of allowed events
+        self.valid_event_types = [constants.VI_EVENT_IO_COMPLETION,
```
MatthieuDartiailh: This could be a set too.

pyvisa-py/highlevel.py
```diff
@@ -93,25 +93,53 @@ def get_debug_info():
...
         self.sessions = {}
 
-    def _register(self, obj):
+        #: map event handle to event object.
+        #: dict[int, Event]
+        self.events = {}
+
+    def _generate_handle(self):
+        """Creates a random but unique handle for a session object,
+        event or find list
```
MatthieuDartiailh: What do you mean by `find list` ?

pyvisa-py/highlevel.py
```diff
@@ -431,9 +470,146 @@ def unlock(self, session):
...
+        :rtype: - eventtype
+                - event
+                - :class:`pyvisa.constants.StatusCode`
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return None, None, StatusCode.error_invalid_object
+
+        # FIXME: for now calling Session's own wait_on_event() is the only way
```
MatthieuDartiailh: I am not sure I follow why this is an issue. Could you elaborate ?

pyvisa-py/highlevel.py
tivek: Before factoring out common functionality to the Session parent class, I am inclined to see some other Session types support events. Do you think different Session subclasses would need fancier filtering, eg. by mechanism and event type?

MatthieuDartiailh: I think there is a misunderstanding here. My point is that the VisaLibrary should not filter the mechanisms but let the session decide if it supports the mechanism.

pyvisa-py/highlevel.py
tivek: A find list is a VISA object used to iterate over available resources: http://zone.ni.com/reference/en-XX/help/370131S-01/ni-visa/vifindrsrc/ As far as I understand the documentation, find list handles share the pool together with handles of resource managers, resource sessions, and events. A session can't have the same handle as an event or a find list because eg. viClose() needs to work for all of these object types.

pyvisa-py/highlevel.py
tivek: Ah, you're correct, sorry. I agree of course, the session should filter itself.

Here I was quite confused about separation of concerns. For instance, I see that pyvisa.VisaLibraryBase already defines a .handlers member instead of having handler registration, storage, mechanism etc. delegated to derived classes.

pyvisa-py/highlevel.py
tivek: This particular comment probably does not belong in there anymore, but the gist of it is I am unclear how to nicely implement the VI_HNDLR mechanism. Event handlers are sitting stored in PyVisaLib.handlers. They are supposed to be called by Sessions (async nonetheless, and that is a whole other can of worms) which should pass event information to them. This was a note for me that the only clean way for sessions to pass events back to the library is to return them from wait_on_event.

pyvisa-py/highlevel.py
MatthieuDartiailh: I guess that it is because pyvisa when using NI-VISA (or similar) does have access to sessions but need to keep track of handlers.

pyvisa-py/highlevel.py
MatthieuDartiailh: Ok in that case I suggest the following:
```python
''''Creates a random and unique handle.

Handles are used for:
- session object (viSession)
- events (viEvent)
- find list (viFindList)
""" 

This way it is easier to find the reference when looking. My confusion was also coming from the fact that pyvisa never expose find lists.

File: pyvisa-py/highlevel.py (hunk 5 above)
Comment (MatthieuDartiailh) [reply pyvisa-py/highlevel.py]: Once again I feel lack I don't have a good enough understanding of VISA events. And of how this would translate in Python (is the ctype handler called in a different thread). If we go further in the direction of event support we should investigate making handler easier to use. In particular I am unclear how to make both backends behave in a similar manner. 
Ideally I would like top be able to pass a python function as handler but that would require some work in the case of the ni backend.

File: pyvisa-py/highlevel.py (hunk 1 above)
Comment (tivek) [reply pyvisa-py/highlevel.py]: Looks that way, yes. I'd say keeping track of handlers is an implementation detail that should go into NIVisaLibrary.

File: pyvisa-py/highlevel.py (hunk 5 above)
Comment (tivek) [reply pyvisa-py/highlevel.py]: I feel implementing event handlers correctly in pyvisa-py is going to be very difficult for little gain. This part of VISA seems to be best implemented at driver level with callbacks into the app.

For example, I have no idea how to have the VI_HNDLR mechanism detect REN line changes of a GPIB board without continuously polling in a separate thread, and even with all the polling a quick pair of on-off events could easily be skipped completely.

I suggest testing and stabilizing the VI_QUEUE mechanism first. It is synchronous and simple. It only needs to implement Session.wait_on_event() which empties the session's own event queue or waits for an event to happen. That alone would already cover most of GPIBInstrument synchronization needs.

Comment (MatthieuDartiailh) [reply pyvisa-py/highlevel.py]: I do agree let's focus on the queue mechanism that is easier. It is just that diving into the pyvisa side of handlers made me realize that they may be some work to there to to make it more usable.

Comment (tivek) [reply pyvisa-py/highlevel.py]: Let me know if I can help on the pyvisa front.

Comment (MatthieuDartiailh) [reply pyvisa-py/highlevel.py]: Currently my main difficulty is to access the hardware required for testing. So if you can help with testing changes it would be great. I have some open PRs both on pyvisa-py and pyvisa that can use a round of testing. 
For the problem of handlers, I need to find time to look more in details how they currently work. Also since people do not generally complain about them, this is rather low priority.
Out of curiosity as one of the few recurrent contributor would you be interested in joining the organization ?

File: pyvisa-py/gpib.py [hunk 6]
```diff
@@ -38,6 +38,46 @@ def _inner(self):
     raise
 
 
+class GPIBEvent(object):
```
MatthieuDartiailh: As I mentionned in my review, I am not sure how to expose this since pyvisa does not expose events. We should perhaps think about how to expose this in pyvisa and how to make it compatible with pyvisa-py.

tivek: `GPIBEvent` is not a user-facing class, it is an implementation detail: `PyVisaLibrary` stores `GPIBEvent` objects in the `PyVisaLibrary.events` dict and passes `any get_attribute()` and `close()` calls down to them. Such "attribute query workers" are required pieces of the puzzle to fully conform to pyvisa's handle-based backend interface.

I agree that pyvisa needs a more pythonic way to expose events, and I think it is actually quite straightforward to get there. Currently, events _are_ actually exposed to the user as raw VISA event contexts/handles. This is not very pythonic but is nevertheless there as a public and documented API. Our user can get hold of an event handle in two main ways:

1. `visalib` calls the installed event handlers and passes the event handle as a parameter,
2. `Resource.wait_on_event()` returns a `WaitResponse` object which exposes a member called `context`.

In both cases, the user can find out more about the event by calling `visalib.get_attribute(handle, attribute)` or close the event using `visalib.close(handle)`.

Instead of low-level handles/contexts, pyvisa should give users some kind of an event object that can be queried and knows how to clean up when it is done. The `WaitResponse` class already fits the bill nicely: all it needs to become user-friendly are `get_attribute()` and `close()` methods and renaming it eg. `Event`. The remaining step is to wrap user-provided event handlers so that they are passed `Event` objects when called instead of raw event handles.

MatthieuDartiailh: Indeed this sounds like a plan.

For the work at hand, even if we do not implement event support right away, we can probably set up a generic class that could be used by all resources rather than an explicitly GPIB restricted class, no ?

tivek: Right. Looks pretty generic actually. If any session needs to compute event attributes on the fly, they can override get_attributes(). Right now I'm struggling with bikeshedding :) I think "Event" should only be the name of the user-facing pyvisa class.

pyvisa-py/highlevel.py
MatthieuDartiailh: I got the proper privileges so I will invite you directly @tivek.

MatthieuDartiailh:
@tivek ping me when this is ready for further reviews.

pyvisa-py/highlevel.py
tivek: > @tivek ping me when this is ready for further reviews.

Right now I feel GPIB event queues are done. It still needs another round of testing and comparing with the NI backend. Right now I only have access to linux-gpib, but during the winter break I will most likely get my hands on a Windows machine with NI hardware.

Apart from that there is the event handler rework on the pyvisa side as discussed above: https://github.com/tivek/pyvisa/tree/expose_events. I should have a WIP pull request soon.

tivek: > I got the proper privileges so I will invite you directly @tivek.

Great, thank you.

tivek:
A quick update. The event work in pyvisa was on hold due to hardware issues. I expect to be able to continue working on it in the following weeks. In the meantime I plan to thoroughly test the changes in pyvisa-py on our test equipment.

MatthieuDartiailh:
Sounds good @tivek . Let me know if I can do anything. I am currently working on pyvisa documentation and I will try to give a shot at implementing GPIB INTFC for pyvisa-py once I am done(see https://github.com/pyvisa/pyvisa/issues/400 for a use case).

MatthieuDartiailh:
@tivek any news on this ?

tivek:
> @tivek any news on this ?

I am trying to replicate the behavior of event queues in NI-VISA. There appear to be some differences in how my KUSB-488A with linux-gpib handles GPIB events with respect to NI adapters on Windows. Not sure if it is down to my code (most likely), the instrument (Tek osci), OS, driver version... This is taking a bit more tinkering than I thought :)

MatthieuDartiailh:
@tivek did you made any progress on this ? I am planning new releases for end of July, beginning of August, would like to push to get this in or will this wait for the next one ?

tivek:
> @tivek did you made any progress on this ? I am planning new releases for end of July, beginning of August, would like to push to get this in or will this wait for the next one ?

Thank you for your patience. I was until recently out of town and without lab access.

I am having increasing difficulties to reproduce behavior of Windows NI-VISA with Linux-GPIB with my instruments. I am not sure if it is something to do with how I initialize the instruments over GPIB or if I am misunderstanding VISA events.

My estimate is GPIB event support will not be ready for the next release. I am planning to exclusively work on this in August when the instruments are not likely to be used for regular experiments.

MatthieuDartiailh:
Sounds good and thanks for the update. Could you also have a look st https://github.com/tivek/gpib_ctypes/issues/3 if you have a moment ?

LongnoseRob:
Was looking for events use with pyvisa-py and found this PR, any info available when work on this will continue?

MatthieuDartiailh:
I do not have the bandwidth at the moment to take over @tivek work, and I did not hear from him since his last post. If this is something you are interested in however, I will be happy to review a PR.

tivek:
> I do not have the bandwidth at the moment to take over @tivek work, and I did not hear from him since his last post. If this is something you are interested in however, I will be happy to review a PR.

Hello,

working from home during the last year has left me without access to gpib equipment so unfortunately I am mostly stuck as far as testing on real-life equipment goes.

Where I got stuck was: the linux-gpib driver for my NI controller either has missing or buggy automatic serial polling. This is needed for any kind of practically useful event support in pyvisa-py. With that bit not really working on my setup, I was not able to make meaningful progress.

Hopefully with vaccinations going forward I'll be able to pinpoint what exactly is the problem and see how to continue.

MatthieuDartiailh:
Sure go ahead

LongnoseRob:
>  Great. Actually, what would help is testing some short gpib code to see if and how autopoll works on your setup. Then we can proceed with rudimentary events. Would that be ok? If so, it would be great if you could find some quick-but-not-too-quick action or measurement on HP3478A which would eg. raise the SRQ bit when it is done.

Yes, we could set the SRQ mask for the data ready feature, and some ranges the conversation rates are very low (only 2-3 readings per second).

If this is would be too fast we still could using single trigger and  GET (group exceute trigger) to time it even slower.

Could you point me to which source to checkout to get ready for the testing?

tivek:
> > Great. Actually, what would help is testing some short gpib code to see if and how autopoll works on your setup. Then we can proceed with rudimentary events. Would that be ok? If so, it would be great if you could find some quick-but-not-too-quick action or measurement on HP3478A which would eg. raise the SRQ bit when it is done.
> 
> Yes, we could set the SRQ mask for the data ready feature, and some ranges the conversation rates are very low (only 2-3 readings per second).
> 
> If this is would be too fast we still could using single trigger and GET (group exceute trigger) to time it even slower.
> 
> Could you point me to which source to checkout to get ready for the testing?

Sounds like a plan.

The code I used for testing sits locally in my lab - sorry about that. In the next one or two days I am going on site and then I can prepare a minimal example for you to test on linux-gpib.

Basically, in order to implement wait_on_srq() for pyvisa-py we need to ibwait() for a RQS event and then check if the device in question generated the request. I observed that linux-gpib with the NI GPIB-USB-HS ignores the IbcAUTOPOLL configuration option and treats it as always disabled, so an ibwait() on the RQS event times out. On the other hand, NI GPIB drivers on Windows respects autopoll as expected.

LongnoseRob:
@tivek just wanted to check if you had a chance to retrieve the code used for testing?

tivek:
> @tivek just wanted to check if you had a chance to retrieve the code used for testing?

@LongnoseRob thanks for your patience, I had to leave it for next week because of some sideffects from the covid shot. Luckily next week I should get more time allotted to play with the instruments.

LongnoseRob:
@tivek 
DId you had a chance to check your sources for the tests?
I think I would be able to do some tests in the next week if I would get some instructions.

hinxx:
Has this PR been abandoned? Will apply against the latest master branch?

MatthieuDartiailh:
@tivek has not given any news recently. If you are interested by this work I think you can pick it up where he left it.

Issue #137, opened 2018-04-08
Document the existence of gpib_ctypes


MatthieuDartiailh:
This may be of interest for Windows users. Of course as the library is still quite young we need to warn the user.

Issue #105, opened 2017-09-22
GPIB Instrument Support for Windows


tkzilla:
Is there a plan to support GPIB instruments on Windows? linux-gpib support is nice, but there are some instruments/software that only run on Windows and they use GPIB loopback addresses.

MatthieuDartiailh:
I am not aware of any library other than visa providing GPIB support on windows, if you do please let us know.

tivek:
GPIB drivers on Windows sit "below" the Visa layer. As far as I know, the GPIB C API exposed by gpib-32.dll/obj on Windows is the same as the C API of linux-gpib. So the question then becomes whether there are Python bindings for GPIB that work both on Linux and Windows.

There is an attempt to build the linux-gpib Python bindings on Windows: https://sourceforge.net/p/linux-gpib/mailman/message/36149135/

Instead of building a Python module in C like linux-gpib does, another way would be to use ctypes and call the GPIB C libraries. The old gpib-devices library does that quite successfully on both Windows and Linux, see [libgpib.py](http://bazaar.launchpad.net/~jordens/gpib-devices/trunk/view/head:/gpib/libgpib.py). However, that is only a straightforward wrap of the C API and some effort is required to have the same interface as linux-gpib.

MatthieuDartiailh:
This sounds very interesting @tivek. Please keep us posted on your progress. As soon as you have something that could be supported by pyvisa-py please let us know and we will try to provide some guidance on how to proceed from there.

tivek:
The rough first commit of linux-gpib-like bindings that also works on Windows is here: https://github.com/tivek/gpib_ctypes

I implemented the low-level `gpib` module from scratch by referring to linux-gpib Python bindings docstrings, NI GPIB C docs as well as linux-gpib docs. However, the nice, object-oriented, capital-letter `Gpib` module that pyvisa-py is using is copied directly from the linux-gpib project. Since linux-gpib is GPL2, that means for the time being gpib_ctypes is given under GPL2.

So far I tried out the code on Windows with NI GPIB-USB-HS, driver version 3.1.0.49154. I plan to also test it on Debian with linux-gpib in the coming days.

How to try it out:
- install pyvisa, pyvisa-py, gpib_ctypes

```python
from gpib_ctypes import make_default_gpib
make_default_gpib() # monkey-patch us to take precedence over eg. linux-gpib's implementation

import visa
rm = visa.ResourceManager('@py')
rm.list_resources()
[...]

PR #176, opened 2018-12-12
use and document gpib-ctypes

tivek:
related to #137 #105

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,9 @@ PyVISA-py Changelog
 0.4.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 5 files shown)


MatthieuDartiailh:
May be worth adding a mention in the docs too.

tivek:
I'll just make one more quick test on Windows and confirm during the day.

tivek:
> May be worth adding a mention in the docs too.

Right. Will do.

tivek:
I'm happy with the change. If you agree, I think it is ready.

MatthieuDartiailh:
I made a minor fix to the docs. Merging.

PR #179, opened 2019-01-15
Fix custom USB timeout and recover after USBTMC timeout

### LINKED ISSUE #177: USBTMC instrument does not recover from timeout; all subsequent commands also time out
I am using pyvisa-py to connect to a Tektronix FCA3000 via USBTMC.
When a read from my instrument hits the timeout, my session with that instrument does not recover.
All subsequent commands and reads also result in timeout.

For example:

    >>> import pyvisa
    >>> rm = pyvisa.ResourceManager("@py")
    >>> instr = rm.open_resource("USB::0x0699::0x3000::379751::INSTR")
    >>> instr.query("*IDN?")
    'TEKTRONIX, FCA3000, 379751, V1.28 25 Aug 2010 11:52\n'
    >>> instr.query("MEAS:FREQ?")  # instrument responds too slow
    ...
    pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
    >>> instr.query("*IDN?")  # instrument should respond immediately
    ...
    pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.

I'm using the latest version of pyvisa-py from Github.
Debian 9.6 - Python 3.5.3 - pyvisa 1.8 - PyUSB 1.0.0 - libusb 1.0.21 - Linux 4.9.110

When I try the same test with NI-VISA, the instrument recovers after timeout and responds quickly and correctly to subsequent commands. (I.e., the first timeout still occurs, but the subsequent IDN command answers quickly.)

Looking at USB traces with Wireshark, I notice that NI-VISA sends an explicit abort sequence after a read timeout, while pyvisa-py does not. The USBTMC spec requires that the host sends this abort sequence when it fails to complete a read operation.

I modified pyvisa-py to add the USBTMC abort sequence, partly based on a similar implementation from python-usbtmc. With that change, my instrument recovers from timeout and answers quickly to subsequent commands.
I will submit a pull request with my changes ...

### LINKED ISSUE #178: USBTMC session does not apply custom timeout
When connecting to an instrument via USBTMC, I find that custom timeout settings are not applied.

For example:

    >>> rm = pyvisa.ResourceManager('@py')
    >>> instr = rm.open_resource("USB::0x1313::0x8078::P0019304::INSTR")
    >>> instr.timeout   # get default timeout
    2000
    >>> instr.read_raw()  # this will surely time out
    ... (after 1 second)
    pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
    >>> instr.timeout = 5000  # set 5 seconds timeout
    >>> instr.read_raw()
    ... (again after 1 second)
    pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.

So it looks like "instr.timeout" is being ignored and a default of 1 second applied instead.

I debugged the timeout handling in pyvisa-py and found a bug in the USBInstrSession class:

        for name in ('SEND_END_EN', 'TERMCHAR', 'TERMCHAR_EN', 'TMO_VALUE'):
            attribute = getattr(constants, 'VI_ATTR_' + name)
            self.attrs[attribute] = attributes.AttributesByID[attribute].default

Direct assignments to `self.attrs[]` accidentally disable the getter/setter functions that are defined for those attributes. As a consequence, timeout changes will not be applied to the USBTMC protocol instance. Instead of directly manipulating `self.attrs`, I think it would be better to call `self.set_attribute()` at this point.

When I make this change, indeed custom timeout settings are correctly applied.
I will submit a pull request ...

jorisvr:
48ef42325506da999e976c0cfe67e79f9bdc1eab implements an USBTMC abort sequence after a read timeout.
Similar abort sequences are used by python-usbtmc and NI-VISA to recover after a read timeout.
Also improves a few specific USBTMC technicalities.
Fixes #177.

cdc3e47912349d08a4eda58eb8676b8c4a4fa442 fixes a bug in custom timeout handling for USB instruments.
Fixes #178.

### DIFF
--- a/pyvisa-py/protocols/usbtmc.py
+++ b/pyvisa-py/protocols/usbtmc.py
@@ -62,6 +62,15 @@ class Request(enum.IntEnum):
     indicator_pulse = 64
 
 
+class UsbTmcStatus(enum.IntEnum):
+    success = 1
+    pending = 2
+    failed = 0x80
+    transfer_not_in_progress = 0x81
+    split_not_in_progress = 0x82
+    split_in_progress = 0x83
+
+
 def find_tmc_devices(vendor=None, product=None, serial_number=None,
                      custom_match=None, **kwargs):
     """Find connected USBTMC devices. See usbutil.find_devices for more info.
@@ -105,12 +114,15 @@ class BulkInMessage(namedtuple('BulkInMessage', 'msgid btag btaginverse '
     @classmethod
     def from_bytes(cls, data):
         msgid, btag, btaginverse = struct.unpack_from('BBBx', data)
-        assert msgid == MsgID.dev_dep_msg_in
+        if msgid != MsgID.dev_dep_msg_in:
+            raise ValueError("Unexpected MsgID 0x{:02x}".format(msgid))
 
         transfer_size, transfer_attributes = struct.unpack_from('<LBxxx', data,
                                                                 4)
 
-        data = data[12:]
+        # Truncate data to the specified length (discard padding).
+        data = data[12:12+transfer_size]
+
         return cls(msgid, btag, btaginverse, transfer_size,
                    transfer_attributes, data)
 
@@ -221,10 +233,7 @@ class USBRaw(object):
         :type data: bytes
         """
 
-        try:
-            return self.usb_send_ep.write(data)
-        except usb.core.USBError as e:
-            raise ValueError(str(e))
+        return self.usb_send_ep.write(data)
 
     def read(self, size):
         """Receive raw bytes to the instrument.
@@ -247,6 +256,7 @@ class USBRaw(object):
 
 class USBTMC(USBRaw):
 
+    # Maximum number of bytes per transfer (for sending and receiving).
     RECV_CHUNK = 1024 ** 2
 
     find_devices = staticmethod(find_tmc_devices)
@@ -291,6 +301,44 @@ class USBTMC(USBRaw):
 
         return interfaces[0]
 
+    def _abort_bulk_in(self, btag):
+        """Request that the device abort a pending Bulk-IN operation."""
+
+        abort_timeout_ms = 5000
+
+        # Send INITIATE_ABORT_BULK_IN.
+        data = self.usb_dev.ctrl_transfer(
+            usb.util.build_request_type(usb.util.CTRL_IN,
+                                        usb.util.CTRL_TYPE_CLASS,
+                                        usb.util.CTRL_RECIPIENT_ENDPOINT),
+            Request.initiate_abort_bulk_in,
+            btag,
+            self.usb_recv_ep.bEndpointAddress,
+            0x0002,
+            timeout=abort_timeout_ms)
+
+        if data[0] != UsbTmcStatus.success:
+            # Abort Bulk-IN failed. Ignore it.
+            return
+
+        # Read remaining data from Bulk-IN endpoint.
+        self.usb_recv_ep.read(self.RECV_CHUNK, abort_timeout_ms)
+
+        # Send CHECK_ABORT_BULK_IN until it completes.
+        for retry in range(100):
+            data = self.usb_dev.ctrl_transfer(
+                usb.util.build_request_type(usb.util.CTRL_IN,
+                                            usb.util.CTRL_TYPE_CLASS,
+                                            usb.util.CTRL_RECIPIENT_ENDPOINT),
+                Request.check_abort_bulk_in_status,
+                0x0000,
+                self.usb_recv_ep.bEndpointAddress,
+                0x0008,
+                timeout=abort_timeout_ms)
+            if data[0] != UsbTmcStatus.pending:
+                break
+            time.sleep(0.05)
+
     def write(self, data):
         """Send raw bytes to the instrument.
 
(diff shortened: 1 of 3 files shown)


pyvisa-py/usb.py
```diff
@@ -194,14 +197,40 @@ def list_resources():
                                   usb_interface_number=intfc))
         return out
 
+    def read(self, count):
```
MatthieuDartiailh: Doing this bypass the whole logic in https://github.com/pyvisa/pyvisa-py/blob/master/pyvisa-py/sessions.py#L436 which I do not really like. Furthermore you simply swallow all usb exceptions that do not match a timeout. I think the proper way to do would be to provide to `self._read` a custom reader function handling the exception and a custom exception class used to differentiate between a timeout and something else. You could define a method/property returning the proper reader function. By default it would simply return `lambda: self.interface.read(count)`, and for USBInstr it would be:
``` python 
def _usb_reader(self):
    """Data reader identifying usb timeout exception.

    """
    try:
        data = self.interface.read(count)
    except usb.USBError as exc:
        if exc.errno in (errno.ETIMEDOUT, -errno.ETIMEDOUT):
            raise USBTimeoutException()
        raise
```
Furthermore it looks to me like this can apply also to all usb session not only USBTMC, which means you can just update read and use a single reader function.

pyvisa-py/protocols/usbtmc.py
```diff
@@ -221,10 +233,7 @@ def write(self, data):
         :type data: bytes
         """
 
-        try:
-            return self.usb_send_ep.write(data)
-        except usb.core.USBError as e:
-            raise ValueError(str(e))
+        return self.usb_send_ep.write(data)
```
MatthieuDartiailh: This method should actually never raise an exception so we should handle this in a more elegant manner. But this can be left another pull request.

jorisvr: Well, it can of course raise an exception for example if somebody pulls the USB plug.
I changed it because I object to the relabeling as ValueError in principle, and because ValueError is objectively the wrong type of exception.

But I understand my version may not be ideal either. If you dislike this change I can drop it.

pyvisa-py/protocols/usbtmc.py
```diff
@@ -291,6 +301,44 @@ def _find_interface(self, dev, setting):
...
+
+        # Send INITIATE_ABORT_BULK_IN.
+        data = self.usb_dev.ctrl_transfer(
+            usb.util.build_request_type(usb.util.CTRL_IN,
+                                        usb.util.CTRL_TYPE_CLASS,
+                                        usb.util.CTRL_RECIPIENT_ENDPOINT),
+            Request.initiate_abort_bulk_in,
+            btag,
+            self.usb_recv_ep.bEndpointAddress,
+            0x0002,
```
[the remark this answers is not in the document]
jorisvr: done

pyvisa-py/protocols/usbtmc.py
```diff
@@ -291,6 +301,44 @@ def _find_interface(self, dev, setting):
...
+        self.usb_recv_ep.read(self.RECV_CHUNK, abort_timeout_ms)
+
+        # Send CHECK_ABORT_BULK_IN until it completes.
+        for retry in range(100):
+            data = self.usb_dev.ctrl_transfer(
+                usb.util.build_request_type(usb.util.CTRL_IN,
+                                            usb.util.CTRL_TYPE_CLASS,
+                                            usb.util.CTRL_RECIPIENT_ENDPOINT),
+                Request.check_abort_bulk_in_status,
+                0x0000,
```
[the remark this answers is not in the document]
jorisvr: done

pyvisa-py/usb.py
jorisvr: Swallowing non-timeout exceptions was an unintentional mistake in my code. I'm glad you caught it.

I opted to bypass Session._read() because it calls the low-level read function repeatedly in a loop, until the result is non-empty. This may be appropriate for raw USB instruments, where you just read until some response or timeout. But USBTMC communication is explicitly message-based, and a device could in principle send an empty response. Repeating the read primitive would not be appropriate in this case.

This is admittedly an edge-case. I think your version would work well enough in practice.

MatthieuDartiailh: I do agree that for a system implementing USBTMC only it would not make sense, but here we are considering system building the VISA layer on top of USBTMC. And generally in VISA when ones performs a read operation the instrument is expected to answer something. Even if the answer is empty I would expect to see the termination character.
Because of this, I still believe it worthwhile to go through the `_read` method to enforce the VISA specification concerning the rule 6.1.

pyvisa-py/protocols/usbtmc.py
MatthieuDartiailh: My point here is that sessions should never leak custom exception but return the proper VISA error code or raise an appropriate Visa exception (the goal is to match the results of using the shared library). But this is not well enforced throughout the code at the moment and fixing it goes beyond the scope of this PR.

MatthieuDartiailh:
Sorry for the slow feedback. See my answer to the read question. Also did you test the changes ? and could you update the changelog ?

pyvisa-py/usb.py
jorisvr: Ok, understood. I implemented your approach with the _usb_reader() function.

pyvisa-py/protocols/usbtmc.py
jorisvr: Clear. Since my change did not really fix it, I put it back the way it was.

jorisvr:
> Sorry for the slow feedback. See my answer to the read question.
No problem. Thanks for your review. I'm glad to have this alternative for NI-VISA, so it's worth putting in some time to improve it.

> Also did you test the changes ?
Yes. I tested simple write/query communication and timeout behaviour with a Thorlabs PM100D power meter and Tektronix FCA3000 frequency counter. Tested again after the last changes.

> and could you update the changelog ?
done.

MatthieuDartiailh:
Thanks for the patch.

Merging

bors r+

PR #180, opened 2019-01-23
fix triggering

greyltc:
This fixes the `assert_trigger` highlevel.py function so that it calls the proper trigger function name in the lower level bits.

Tested with gpib. Should also fix other protocols that use triggering too, like tcpip


MatthieuDartiailh:
Good catch !

bors r+

MatthieuDartiailh:
I should have asked you to update the release notes .....

PR #194, opened 2019-06-26
Add support GPIB INTFC

MatthieuDartiailh:
This re-organizes the code that, before, was heavily mixing INSTR and INTFC capabilities. It also patches gpib_ctypes to provide some extra functions we need. https://github.com/tivek/gpib_ctypes/issues/3 tracks the issue that should also be discussed in linux-gpib.

This code has not been tested yet. Anybody that can help in testing that code is very welcome to do so. Note that to access all the new functionalities one needs gpib_ctypes since linux-gpib does not expose some functions to Python. The typical use case for those is described in: https://github.com/pyvisa/pyvisa/issues/400.

@jondoesntgit you may be interested in this. I did not implement the buffer_read/write mostly because I am not sure they are necessary. Using read should work (you have to go through the VISA library for the time being but I would be interested in discussing adding it to the resource since it is supposed to support it.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -354,9 +354,8 @@ class _GPIBCommon(object):
                 return constants.VI_FALSE, StatusCode.success
 
         elif attribute == constants.VI_ATTR_SEND_END_EN:
-            # replace IbaEndBitIsNormal 0x1a
-	    # IbcEndBitIsNormal relates to EOI on read()
-            # not write(). see issue #196
+            # Do not use IbaEndBitIsNormal 0x1a which relates to EOI on read() 
+            # not write(). see issue #196 
             # IbcEOT 0x4
             if ifc.ask(4):
                 return constants.VI_TRUE, StatusCode.success
@@ -424,9 +423,8 @@ class _GPIBCommon(object):
                 return StatusCode.error_nonsupported_attribute_state
 
         elif attribute == constants.VI_ATTR_SEND_END_EN:
-            # replace IbaEndBitIsNormal 0x1a
-	    # IbcEndBitIsNormal relates to EOI on read()
-            # not write()
+            # Do not use IbaEndBitIsNormal 0x1a which relates to EOI on read() 
+            # not write(). see issue #196 
             # IbcEOT 0x4
             if isinstance(attribute_state, int):
                 ifc.config(4, attribute_state)
@@ -503,7 +501,6 @@ class GPIBInterface(_GPIBCommon, Session):
         return ['GPIB0::%d::INTFC' % pad for pad in _find_listeners()]
 
     def after_parsing(self):
-        logger.debug("PARSED: ", self.parsed)
         minor = int(self.parsed.board)
         sad = 0
         timeout = 13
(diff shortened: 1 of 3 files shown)


jondoesntgit:
Very nice.

I’m away from the lab right now on an internship and I moved away from the su lotion that I was investigating before with using INTFC towards solving the problem another way. However, if the need arises again in the future, I’ll revisit this.

dirkenstein:
@MatthieuDartiailh - I'm really struggling with this.

Does it need a custom gpib_ctypes version to work?

The pip included one doesn't export _lib and i have no idea how to override the module import rules from another package. Is there a newer one? 

If i modify gpib_ctypes to export _lib it loads but complains about:

TypeError: Can't instantiate abstract class GPIBSession with abstract methods _get_attribute, _set_attribute, close

MatthieuDartiailh:
Seems you figured it out before I had a chance to answer...

MatthieuDartiailh:
Ok so I fixed the import for gpib ctypes. However I could not test in great details:
- the machine I have access to has both NI and Keysight 
- I am using a Keysight gpib adpater
- I could not manage to get gpib-ctypes to load the proper gpib library (I believe at least)

I was planning to use the following test script with a SRS 830. It works using VISA, if somebody can test pyvisa-py it would be great. @dirkenstein I tried reading from the interface but it was not working so I ended reading from the instrument. Is that normal ? How did you manage to read from the interface in your usecase ?
```python
from visa import log_to_screen
from pyvisa import ResourceManager
from pyvisa.highlevel import list_backends

BACKEND = '@py'

GPIB_ADDRESS = 8

log_to_screen()

print(list_backends())

print('Opening resource manager')
rm = ResourceManager(BACKEND)
print(rm)

print('Opening resource')
resource = rm.open_resource(f'GPIB1::{GPIB_ADDRESS}::INSTR', open_timeout=1)
resource.timeout = 5
resource.read_termination = '\n'
intfc = rm.open_resource('GPIB1::INTFC')

print('Querying ID')
print(resource.query('*IDN?'))

# Set the sampling rate to 2 Hz
resource.write('SRAT 20')

resource.write('FAST 2')
resource.write('STRD')
intfc.send_command(f'MLA TALK {GPIB_ADDRESS}')

try:
    while True:
        print(resource.read_bytes(4))
finally:
    resource.write('FAST 0')
```

dirkenstein:
@MatthieuDartiailh  I was using `intfc.visalib.buffer_read(intfc.session, 3)` to read 3 bytes from the bus. Never tried reading directly on the interface. I think that requires modification to pyvisa to support.

dirkenstein:
Just realized this doesn't work as I'd hoped:

1) `buffer_read()` is not implemented.
2) `read()` assumes it will have an existing interface

MatthieuDartiailh:
Wow looking at your code example for `send_command` made me realize how off my own code was and as a consequence I have no idea why it do not error out and gave something that looked correct with a Keysight adapter (but the same with NI adapter was failing). I also realized that the values for MLA and MTA are nowhere in pyvisa which is something we could/should fix. 
I did not implement `read_buffer` cause I was not sure it was truly useful (compare to read). I pushed a change to allow to read/write on an interface. I won't have access to the hardware to test this till at least mid-August, so this won't be included in the next release, but if you can get it to work it would be great !
Furthermore, I am pretty sure that the attributes are currently a mess and I will need some time to figure them out properly.

MatthieuDartiailh:
If you can confirm this works, I may find the time to push this in the next release  (would be a great feature to have).

MatthieuDartiailh:
@dirkenstein I think this is good to go now. I am out the lab currently so if you can cross-check my last changes, I will merge and proceed to the release. Thanks A LOT for your support with.

MatthieuDartiailh:
@dirkeinstein I would like to merge this week so if you can have a last look it would be great.

PR #197, opened 2019-07-14
Updated GPIB interface for review

dirkenstein:


pyvisa-py/gpib.py
```diff
@@ -165,7 +165,8 @@ def _get_timeout(self, attribute):
             else:
                 # value is 0 or out of range -> infinite
                 self.timeout = None
-        return super(GPIBSession, self)._get_timeout(attribute)
+        return super(_GPIBCommon, self)._get_timeout(attribute)
+        #return self.timeout
```
MatthieuDartiailh: Leftover ?

pyvisa-py/gpib.py
```diff
@@ -337,21 +344,22 @@ def _get_attribute(self, attribute):
...
-            if self.interface.ask(27):
+            if ifc.ask(27):
                 return constants.VI_TRUE, StatusCode.success
             else:
                 return constants.VI_FALSE, StatusCode.success
 
         elif attribute == constants.VI_ATTR_SEND_END_EN:
-            # IbaEndBitIsNormal 0x1a
-            if self.interface.ask(26):
+            # replace IbaEndBitIsNormal 0x1a
```
MatthieuDartiailh: Could you add a more detailed discussion here ? and link to the github issue ?

dirkenstein:
Added some comments as a response

MatthieuDartiailh:
Great I will merge during the week as I test.

dirkenstein:
Hi,

I need to release an extra update-

1) gpib_command definition is wrong- the visa implementation takes byes as
an argument not an int.
The pyvisa-py documentation contradicts the pyvisa documentation in that
respect.

There are two outstanding issues:

1) The updates won't work with gpib_ctypes unless you allow _lib to be
exported in __init__.py of that package. I can't figure out any way to work
around this. I suggest we add the missing functions directly to
gpib_ctypes. Even though this breaks complete compatibility with linux-gpib
it makes sense to extend the ctypes version as at least that seems actively
maintained and is designed to also work with other 488.2 libraries.

1) The pyvisa-py implementation of gpib_command is based on the gpib_ctypes
version which should work. It looks like the linux-gpib version is missing
the length argument so is actually buggy.

Regards,

Dirk

MatthieuDartiailh:
Thanks for the fix I will merge now. I have, I believe, a workaround to access _lib (may not be pretty though). I am in contact with the author of gpib_ctypes and I do hope to upstream those changes for sure.

PR #198, opened 2019-07-20
Fixes to handle interface reads and writes

dirkenstein:
Fixed calls to directly imported library functions to use gpib_lib
Added support for buffer_read()/buffer_write() functions (just call read() and write() for now)
Fixed support  for read() and write() on interface objects

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -12,6 +12,7 @@
 
 from __future__ import division, unicode_literals, print_function, absolute_import
 from bisect import bisect
+import ctypes  # Used for ibln not ideal
 
 from pyvisa import constants, logger, attributes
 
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Sorry I missed your changes before I pushed some of mines. Why did you remove the extra function from gpib_ctypes ? Those are still absent from gpib_ctypes. I used the `or` idiom to shorten a bit the code, and missed the atn changes. So you can probably rebase and I will merge.

dirkenstein:
Don't panic- i think the missing function was an error on my part- i will rebase once i work out exactly what I did.

dirkenstein:
this one should merge OK now- the reason I removed the missing gpib_ctypes functions was that i had an updated gpib_ctypes version ready with all the  missing functions implemented. 

Then i realized that there hadn't been any updates to gpib_ctypes for a while so I didn't want to issue a pull request for it- probably just me being nervous.

I needed the buffer_read() and buffer_write() functions because i was using them in my native-visa pyvisa codebase.

MatthieuDartiailh:
Could you please configure your editor to trim trailing white spaces ? Also there are a couple of weird indentation/blanlkline changes.

Please open the PR on gpib_ctypes, I will try to push on the author. That kind of library is rather low maintenance hence the low traffic.

Finally we can add buffer_read/write, but I would prefer to understand it properly first to do it correctly. For your use case, I believe that read should be enough.

MatthieuDartiailh:
Thanks. Looks good.

PR #199, opened 2019-07-21
Fixed exception on interface open caused by missing primary address

dirkenstein:
This got missed when after_parsing was removed from the interface class


MatthieuDartiailh:
Don't we also need `self.interface =None` for the interface case ?

dirkenstein:
Seemed to work without it, but you're probably right.

MatthieuDartiailh:
Thanks.  And thanks for putting up with my poorly debugged code. Can you reproduce your application in pyvisa-py now ?

PR #200, opened 2019-07-22
Removed SEND_END_EN from variable parameters

dirkenstein:
This was hard to track down- adding `VI_ATTR_SEND_END_EN` to `self.attrs` disables attribute setting using` _set_attribute()` as it's now just a variable.

You can't set defaults this way except for attributes that are not implemented in `_set_attribute()`.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -159,7 +159,7 @@ class _GPIBCommon(object):
         self.set_attribute(constants.VI_ATTR_TMO_VALUE,
                            attributes.AttributesByID[constants.VI_ATTR_TMO_VALUE].default)
 
-        for name in ('SEND_END_EN', 'TERMCHAR', 'TERMCHAR_EN'):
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
So sorry for this ! I just copy pasted what was done on other resource because I realized we were missing termination characters. As I said earlier I need to look at the VISA spec to clean up the attributes. I am pretty sure that we cannot share all of them between INSTR and INTFC. Thanks for tracking this down.

dirkenstein:
Don't panic- my application works now on both Keysight VISA and pyvisa-py
with linux-gpib-4.2.0.
The only difference is now down to the Agilent 82357 not supporting
iblines() correctly on linux-gpib, so list_resources() doesn't allow
auto-discovery.
Nothing to do with pyvisa or pyvisa-py.

Dirk

MatthieuDartiailh:
WOW that's great to hear @dirkenstein ! Thanks a lot for taking the time to debug this. I will try to clean up the attribute mess and push this into the release.

PR #201, opened 2019-07-28
Adds functionality for multiple boards in one system, fixes lockup on scan

dirkenstein:
Supports up to 16 boards as per linux-gpib.

Add functionality to never perform a `listener()` on the controller address. Previously it would not yield  the address value but still  call `listener()`. This causes some gpib interfaces to lock up with a bus error until an IFC is issued. May be related to whether we call MTA of the controller as part of `ibln()` or not but some boards need that. 

Note the issue with MTA seems to be specific to linux-gpib. Testing with NI488.2 will be performed to make sure nothing breaks.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -539,7 +539,7 @@ class GPIBInterface(_GPIBCommon, Session):
         try:
             return self.controller.command(command_bytes), StatusCode.success
         except gpib.GpibError as e:
-            return convert_gpib_error(e, self.interface.ibsta(), 'gpib command')
+            return convert_gpib_error(e, self.controller.ibsta(), 'gpib command')
 
     def gpib_send_ifc(self):
         """Pulse the interface clear line (IFC) for at least 100 microseconds.
@@ -555,7 +555,7 @@ class GPIBInterface(_GPIBCommon, Session):
             self.controller.interface_clear()
             return StatusCode.success
         except gpib.GpibError as e:
-            return convert_gpib_error(e, self.interface.ibsta(), 'send IFC')
+            return convert_gpib_error(e, self.controller.ibsta(), 'send IFC')
 
     def gpib_control_atn(self, mode):
         """Specifies the state of the ATN line and the local active controller state.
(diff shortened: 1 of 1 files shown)


pyvisa-py/gpib.py
```diff
@@ -126,7 +135,14 @@ def convert_gpib_error(error, status, operation):
         else:
             return constants.StatusCode.error_system_error
 
-
+def convert_gpib_status(status):
```
MatthieuDartiailh: This seems to be unused. If it is true please remove it.

pyvisa-py/gpib.py
```diff
@@ -68,16 +68,25 @@ def _inner(self):
 # TODO: Check board indices other than 0.
 BOARD = 0
```
MatthieuDartiailh: This is probably not needed anymore. Could you please remove it and any reference to it.

pyvisa-py/gpib.py
```diff
@@ -507,7 +523,7 @@ class GPIBInterface(_GPIBCommon, Session):
 
     @staticmethod
     def list_resources():
-        return ['GPIB0::%d::INTFC' % pad for pad in _find_listeners()]
+        return ['GPIB%d::INTFC' % board for board, pad, in _find_boards()]
```
MatthieuDartiailh: The last comma in `board, pad,` is useless and should be removed for style uniformity.

pyvisa-py/gpib.py
```diff
@@ -522,8 +538,8 @@ def gpib_command(self, command_bytes):
         """
         try:
             return self.controller.command(command_bytes), StatusCode.success
-        except gpib.GpibError:
-            return 0, convert_gpib_status(self.interface.ibsta())
+        except gpib.GpibError as e:
+            return convert_gpib_error(e, self.controller.ibsta(), 'gpib command')
```
MatthieuDartiailh: Good catch !

MatthieuDartiailh:
I merged because I wanted to make sure not to break too many things when updating the handling of attributes. If you can test the new attributes. If you are happy with it we can probably merge #194 and include it in the next release.

PR #203, opened 2019-08-08
Removed spurious comma and BOARD variable

dirkenstein:
Note the convert_gpib_status function is actually used.

### DIFF
--- a/pyvisa-py/gpib.py
+++ b/pyvisa-py/gpib.py
@@ -65,9 +65,6 @@ def _patch_Gpib():
 _patch_Gpib()
 
 
-# TODO: Check board indices other than 0.
-BOARD = 0
-
 def _find_boards():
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
I am confused I fixed those as part of the attribute cleanup as can be seen here https://github.com/pyvisa/pyvisa-py/blob/c9fdae8bbc811bbcc8fc2db2b8344bfa31e108f5/pyvisa-py/gpib.py. I will close because it does not seem relevant anymore (I don't get how Github computes the diff here).

dirkenstein:
OK- if you fixed those i missed one of your earlier diffs somehow

MatthieuDartiailh:
It is part of my last commit. Hopefully know everything works as expected, if you can confirm it is so we will be able to merge soon.

PR #205, opened 2019-08-23
Serial: Properly set stop bits

### LINKED ISSUE #204: Serial: Setting stop bits does nothing
With a serial instrument `dev`, the following code should set 2 stop bits
```python
from pyvisa import constants
dev.stop_bits = constants.StopBits.two
print(dev.stop_bits)
```
However, the print statement outputs `StopBits.one`.

This occurs with pyvisa-1.10.0 and pyvisa-py-0.3.1. It also occurs with pyvisa-1.10.0 and the latest master of pyvisa-py.

circuitfox:
Fixes #204

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.4.0 (unreleased)
 ------------------
 
+- fix not setting stop bits with serial devices PR #205
 - fix improper wait time before a timeout in the TCPIP backend PR # 173
 - add GPIB support for proprietary device drivers on Windows and Linux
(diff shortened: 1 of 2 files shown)


MatthieuDartiailh:
Thanks for the fix. It looks all reasonable. I assume you tested it and it fixes your problem, right ? 

Could you please update the releasenotes and I will merge ?

circuitfox:
Yes. This was tested with an Agilent 33120A which requires 2 stop bits and was giving errors when that wasn't the case. This fixed that problem.

MatthieuDartiailh:
Great ! Thanks for your contribution.

bors r+

PR #212, opened 2019-10-18
adding usb_control_in functionality

Zanobos:
As discussed here: https://github.com/pyvisa/pyvisa/issues/466 the pyvisa-py backend is missing the implementation of usb_control_in. This patch applies the modification needed to have it working. Tested and working with a bk8600.

Disclaimer: Is implemented the feature of control_transfer, but not the one of optional data stage - I miss the specifications and the knowledge to do that part (I tried with a plain self.read(length), but it's not enough).
If anyone can lead me to the right way, I can provide implementation for that part too.

### DIFF
--- a/pyvisa-py/highlevel.py
+++ b/pyvisa-py/highlevel.py
@@ -510,3 +510,30 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
     def discard_events(self, session, event_type, mechanism):
         # TODO: implement this for GPIB finalization
         pass
+
+    def usb_control_in(self, session, request_type_bitmap_field, request_id, request_value,
+                       index, length=0):
+        """Performs a USB control pipe transfer from the device.
+
+        Corresponds to viUsbControlIn function of the VISA library.
+
+        :param session: Unique logical identifier to a session.
+        :param request_type_bitmap_field: bmRequestType parameter of the setup stage of a USB control transfer.
+        :param request_id: bRequest parameter of the setup stage of a USB control transfer.
+        :param request_value: wValue parameter of the setup stage of a USB control transfer.
+        :param index: wIndex parameter of the setup stage of a USB control transfer.
+                      This is usually the index of the interface or endpoint.
+        :param length: wLength parameter of the setup stage of a USB control transfer.
+                       This value also specifies the size of the data buffer to receive the data from the
+                       optional data stage of the control transfer.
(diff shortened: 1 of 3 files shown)


pyvisa-py/highlevel.py
```diff
@@ -510,3 +510,30 @@ def disable_event(self, session, event_type, mechanism):
...
+                 - return value of the library call.
+        :rtype: - bytes
+                - :class:`pyvisa.constants.StatusCode`
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return StatusCode.error_invalid_object
+        
+        return sess.control_transfer(request_type_bitmap_field, request_id, request_value, index, length)
```
MatthieuDartiailh: Missing blank line at the end of the file.

pyvisa-py/protocols/usbtmc.py
```diff
@@ -431,3 +430,30 @@ def read(self, size):
...
+        """
+
+        self.usb_dev.ctrl_transfer(request_type_bitmap_field,
+                request_id,
+                request_value,
+                index,
+                length,
+                timeout=self.timeout)
+
+        #TODO satisfy the second half of the problem -i.e. sending buffer the optional data stage
```
MatthieuDartiailh: You can try using to do a direct read. Using `self.read` means you are starting a full USBTMC request which is not what we want here. See https://github.com/pyvisa/pyvisa-py/blob/master/pyvisa-py/usb.py#L108

pyvisa-py/protocols/usbtmc.py
```diff
@@ -431,3 +430,30 @@ def read(self, size):
...
+        # return self.read(length) does not work
+        return bytes()
```
MatthieuDartiailh: Missing blank line

MatthieuDartiailh:
ping @Zanobos

Zanobos:
Hi @MatthieuDartiailh ,
I'll try to work on this branch as soon as possible.

pyvisa-py/protocols/usbtmc.py
Zanobos: Resolved in last commit

pyvisa-py/highlevel.py
Zanobos: resolved in last commit

PR #221, opened 2020-03-10
Moved the metadata into setup.cfg

KOLANICH:


MatthieuDartiailh:
I rebased but could not force push. I will open a new PR. Thanks for your contribution (Note that in order for setup.py to behave as expected one needs to set pass use_scm_version otherwise non pip install or editable install miss the version).

Issue #225, opened 2020-05-02
Add support for LAN device name in TCPIP INSTR resources


MatthieuDartiailh:
As reported in https://github.com/pyvisa/pyvisa/issues/508 support for lan device name would enable some user extra user cases.

garlick:
Using wireshark to watch the vxi11 protocol exchange, it looks like the lan device name required by my GPIB ethernet gateway (e.g. `gpib0,4` or `gpib0,9,3`) passes through OK, but after establishing the link, _device_write_ fails.  I noticed that the flag `termcharset` (0x80) is set in the _device_write_ RPC, but the VXI-11 spec says:

> termchrset(80): This flag is set to one(1) if a termination character is specified on a read. The
actual termination character itself is passed in the termchr parameter. This flag is only valid for
device_read.

(B.5.3. Operation Flags, pg 32)
https://www.icselect.com/vxi_spec.html

So maybe the protocol implementation is wrong, and this device is more pedantic about it than others?

I tried modifying `TCPIPInstrSession.write()` to skip setting this flag and the _device_write_ succeeds.
```diff
--- tcpip.py.bak	2020-05-02 13:28:15.291195706 -0700
+++ tcpip.py	2020-05-03 07:14:17.939321057 -0700
@@ -164,10 +164,7 @@
         chunk_size = 1024
 
         try:
-            if send_end:
-                flags = vxi11.OP_FLAG_TERMCHAR_SET
-            else:
-                flags = 0
+            flags = 0
 
             num = len(data)
             offset = 0
```

However, then things go wrong processing the data in the response:
```
Traceback (most recent call last):
  File "./testprog.py", line 25, in <module>
    print(inst.query("*IDN?"))
  File "/home/garlick/.local/lib/python3.6/site-packages/pyvisa/resources/messagebased.py", line 613, in query
    return self.read()
  File "/home/garlick/.local/lib/python3.6/site-packages/pyvisa/resources/messagebased.py", line 427, in read
    message = self._read_raw().decode(enco)
  File "/home/garlick/.local/lib/python3.6/site-packages/pyvisa/resources/messagebased.py", line 400, in _read_raw
    chunk, status = self.visalib.read(self.session, size)
  File "/home/garlick/.local/lib/python3.6/site-packages/pyvisa-py/highlevel.py", line 345, in read
    ret = self.sessions[session].read(count)
[...]

PR #226, opened 2020-05-03
address issues getting ICS 8065 ethernet-gpib gateway working

garlick:
This fixes some issues raised in #225, while sorting out a setup involving an ICS 8065 ethernet-GPIB gateway.  One fix corrects the VXI-11 protocol implementation, and another avoids raising an exception when the 8065 violates RFC 1014 - harmlessly I think;  the glibc XDR implementation doesn't complain about it anyway.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,10 @@ PyVISA-py Changelog
 0.4.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
Thanks for the PR. This looks good. I am going to merge and since a release is long overdue I will release right after.

bors r+

PR #232, opened 2020-05-26
correct import

bollenn:
correct import of struct package


MatthieuDartiailh:
Good catch ! And sorry for the sloppy edit !

bors r+

PR #233, opened 2020-05-26
correct const name

bollenn:
correct const name


bollenn:
@MatthieuDartiailh do you think we can do a patch release including this and https://github.com/pyvisa/pyvisa-py/pull/232? that would help me big time...

MatthieuDartiailh:
The master of pyvisa-py currently requires the master branch of pyvisa. The issue in #232 was introduced after the last release to make pyvisa-py supports pyvisa master which dropped support for Python 2. However this bug does affect 0.4.0. I will backport this to 0.4.0 and release a bugfix release (since #232 does not apply to 0.4.0 there is no need to backport it). Hopefully I will have time to take care of that tomorrow evening.

Issue #227, opened 2020-05-04
[COM] Communication issue with Ansritsu MG37022A using TCPIP, infinite loop in open_resource


Schneider1:
Instrument details
------------------
* Model: Anritsu MG37022A
* Communication: TCPIP
* Link to the documentation (if available): https://dl.cdn-anritsu.com/en-us/test-measurement/files/Manuals/Programming-Manual/10370-10371.pdf

Output of `pyvisa-info`
-----------------------
```
Machine Details:
   Platform ID:    Linux-5.4.0-28-generic-x86_64-with-glibc2.29
   Processor:      x86_64

Python:
   Implementation: CPython
   Executable:     /usr/bin/python3
   Version:        3.8.2
   Compiler:       GCC 9.3.0
   Bits:           64bit
   Build:          Apr 27 2020 15:53:34 (#default)
   Unicode:        UCS4

PyVISA Version: 1.10.1

Backends:
   ni:
      Version: 1.10.1 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.4.0.dev0
      ASRL INSTR:
         Please install PySerial (>=3.0) to use this resource type.
         No module named 'serial'
      USB INSTR:
         Please install PyUSB to use this resource type.
         No module named 'usb'
      USB RAW:
         Please install PyUSB to use this resource type.
         No module named 'usb'
      TCPIP INSTR: Available
      TCPIP SOCKET: Available
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of funcionality.
         No module named 'gpib'
```

Opening the connection with `rm.open_resource('TCPIP0::10.22.197.10')` gets stuck in an infinite loop, without throwing a timeout.

After some search, I found that pyvisa-py keeps waiting for a second segment, since the first segment has the "last" flag disabled. However, there is no next segment sent. So this looks like a bug in the device, BUT when I try the same using the NI-VISA backend, everything works well. In the latter case, I checked with Wireshark that device again sends the "last=False" flag (resulting in a RPC Fragment), but NI-VISA can shomehow cope with the situation.

If I do a very dirty fix and set` last=True `in protocols/rpc.py:377, everything works fine and all other commands set the "last" flag correctly.

Any ideas, why this device behaves so strange? Why does NI-VISA not have a problem with that? How can I solve this?

[...]

PR #242, opened 2020-07-17
Add expected number of packages in rpc to avoid timeout

Schneider1:
As described in #227 some instruments (found it for other Anritsu instruments as well) wrongly set the "next" flag in their response, although no next package follows. This PR introduces a parameter for the expected number of packages, after which `_recvrecord()` can return. For `get_port`, this is set to 1.

Tested and works for me.


MatthieuDartiailh:
I will review in more detail later bu this looks good. I feel like we should refactor the tcpip:instr support since the more we go the less clean is the separation between the pure RPC part and the protocol, but we need test before going that way.

Once again the CI do not trigger this is annoying....

bors try

MatthieuDartiailh:
bors try

MatthieuDartiailh:
Can you run black on your changes ? I did not yet add a pull request template specifying this.

Schneider1:
Sorry for that, I hope now it fits

I guess I can't trigger bors?

bors try

MatthieuDartiailh:
bors try

MatthieuDartiailh:
Looking all good. Thanks for your contribution.

bors r+

PR #244, opened 2020-07-23
fix crash on Windows when opening multiple devices

itayperl:
For some reason, PyUSB on Windows is unable to read the serial_number
attribute from a device that's already open in another pyusb instance,
and crashes with a cryptic "device has no langid" exception when it's accessed.

When a USB resource is opened, PyVISA-py tries to find the device by serial
number. If there's already another open device, its serial_number
attribute will be accessed in this search, triggering the langid
exception.

With this fix we will silently ignore a device when accessing a search
attribute on it raises an exception.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ PyVISA-py Changelog
 - drop Python 2 support and run against PyVISA 1.11 PR #222
 - usbtmc: improve support for USB488 devices. PR #241
   For instrument that support REN_CONTROL, we now always assert the REN line.
+- fix a crash on Windows when opening multiple USBTMC devices
 
 0.4.1 (2020-05-27)
 ------------------
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -62,6 +62,9 @@ to do it http://ask.xmodulo.com/change-usb-device-permission-linux.html.
 On Windows, you may have to uninstall the USBTMC specific driver installed by
 Windows and re-install a generic driver.
 
+Note that on Windows, devices that are already open cannot be detected and will
+not be returned by ``ResourceManager.list_resources``.
+
 Another useful reference for how to configure your system is h
 ttps://github.com/python-ivi/python-usbtmc.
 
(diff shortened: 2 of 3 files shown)


MatthieuDartiailh:
That is bound to confuse people a lot ! Could you add something in the documentation about it.

itayperl:
Hmm, what do you find confusing here? It feels reasonable to me to skip devices that we are unable to query.

The only thing that is a bit confusing to me is that on Linux ResourceManager.list_resources() returns all connected devices, whereas on Windows it only returns devices that are not already open (that's already the behavior today, not affected by this PR).

MatthieuDartiailh:
I am pretty that some people will report that their device is not detected on Windows and as a consequence Ithink it is worth improving the documentation to mention that caveat. I am fine with your implementation.

MatthieuDartiailh:
But we may want to log the fact that specific instrument could not be accessed and if on windows suggest one possible reason.

itayperl:
I added a note about list_resources - is there any other scenario that should be documented?
I don't think we need another log for this case - there's already a warning log [here](https://github.com/pyvisa/pyvisa-py/blob/master/pyvisa-py/usb.py#L223) for this failure in list_resources.

MatthieuDartiailh:
My bad I forgot about that warning.

pyvisa-py/protocols/usbutil.py
```diff
@@ -205,7 +205,11 @@ def cm(dev):
             if custom_match is not None and not custom_match(dev):
                 return False
             for attr, pattern in attrs.items():
-                if not fnmatch(getattr(dev, attr).lower(), pattern.lower()):
+                try:
+                    value = getattr(dev, attr)
+                except Exception:
```
MatthieuDartiailh, line 210: Could we use a more specific error here since we know how this could fail ? (and if it fails for a different reason it may be nice to know about it)

itayperl: OK, changed it to be consistent with the check in usb.py.

MatthieuDartiailh:
#238 has the fix for the format issue. If you are fine waiting I will try to push https://github.com/pyvisa/pyvisa/pull/525 and #238 through next week. If you need the fix on master faster let me know.

itayperl:
No rush! I'll use my fork in the meantime. Thanks!

MatthieuDartiailh:
@itayperl can you address the conflicts ?

itayperl:
> @itayperl can you address the conflicts ?

Done.

PR #269, opened 2020-09-23
serial: trim COM from serial ports and add it when opening

### LINKED ISSUE #267: 'Resource' object has no attribute 'write'
Hi,
I've just updated to the latest version of pyvisa and now get the following error:

```
Python 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyvisa
>>> rm = pyvisa.ResourceManager()
>>> rm.list_resources()
('ASRLCOM3::INSTR', 'ASRLCOM4::INSTR')
>>> my_instrument = rm.open_resource('ASRLCOM3::INSTR')
>>> print(my_instrument.query('*IDN?'))
AttributeError: 'Resource' object has no attribute 'query'
```

I've noted this unusual output:
```
>>> print(my_instrument)
Resource at ASRLCOM3::INSTR

```
expected behaviour on readthedocs is along the lines of:
`<GPIBInstrument('GPIB::14')>`

*pyvisa-info*
PS C:\Users\mainuser\Documents\Python\sm1800> pyvisa-info
Machine Details:
   Platform ID:    Windows-10-10.0.19041-SP0
   Processor:      Intel64 Family 6 Model 142 Stepping 9, GenuineIntel

Python:
   Implementation: CPython
   Executable:     c:\users\mainuser\appdata\local\programs\python\python38\python.exe
   Version:        3.8.3
   Compiler:       MSC v.1924 64 bit (AMD64)
   Bits:           64bit
   Build:          May 13 2020 22:37:02 (#tags/v3.8.3:6f8c832)
   Unicode:        UCS4

PyVISA Version: 1.11.0

Backends:
   ivi:
      Version: 1.11.0 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.5.0
      USB INSTR:
         No module named 'usb'
      USB RAW:
         No module named 'usb'
      TCPIP INSTR: Available
      TCPIP SOCKET: Available
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of funcionality.
         No module named 'gpib'

MatthieuDartiailh:
- [x] Closes #267

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.5.1 (unreleased)
 ------------------
 
+- list serial resources under Windows without the COM prefix #269
 - fix return value of USB close method PR #265
 - fix pyvisa version contraint PR #268
 
--- a/pyvisa_py/serial.py
+++ b/pyvisa_py/serial.py
@@ -6,6 +6,7 @@
 :license: MIT, see LICENSE for more details.
 
 """
+import sys
 from typing import Any, List, Optional, Tuple
 
 from pyvisa import attributes, constants, logger, rname
@@ -29,6 +30,8 @@ except ImportError as e:
         "Please install PySerial (>=3.0) to use this resource type.\n%s" % e,
     )
     raise
+
+IS_WIN = sys.platform == "win32"
 
 
 def iter_bytes(data: bytes, mask: Optional[int] = None, send_end: bool = False):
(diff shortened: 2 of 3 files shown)


MatthieuDartiailh:
Will fail till https://github.com/pyvisa/pyvisa/pull/545 is merged

MatthieuDartiailh:
Actually this revealed a more complicated issue. According to the VISA standard we only expect integers after the resource type (i.e. ASRL1::INSTR) and I started enforcing this in PyVISA, but it appears it will break serial on pyvisa-py both on Windows (but there we can easily fix it) and Linux where it is more complicated since the port can look like a path /dev/ttyAMA0. 

I need to think a bit about how to fix this issue. The windows patch still makes sense to get a more homogeneous experience, but it won't be sufficient for Linux.

MatthieuDartiailh:
I went with the option of allowing ResourceInfo board to be None which allows to preserve all previous and past functionality. This should be good to go.

PR #277, opened 2020-09-28
serial: fix writing to serial resources

### LINKED ISSUE #272: sending ints into pyserial's write() function is not good?

MatthieuDartiailh:
- [x] Closes #272 

### DIFF
--- a/pyvisa_py/serial.py
+++ b/pyvisa_py/serial.py
@@ -178,14 +178,12 @@ class SerialSession(Session):
 
         """
         logger.debug("Serial.write %r" % data)
-        # TODO: How to deal with VI_ATTR_TERMCHAR_EN
         end_out, _ = self.get_attribute(ResourceAttribute.asrl_end_out)
         send_end, _ = self.get_attribute(ResourceAttribute.send_end_enabled)
 
-        # For end_out in (SerialTermination.none, SerialTermination.termination_break)
-        # we can send the data as passed.
-
-        if end_out == SerialTermination.last_bit:
+        if end_out in (SerialTermination.none, SerialTermination.termination_break):
+            pass
+        elif end_out == SerialTermination.last_bit:
             last_bit, _ = self.get_attribute(ResourceAttribute.asrl_data_bits)
             mask = 1 << (last_bit - 1)
             data = bytes(iter_bytes(data, mask, send_end))
--- a/pyvisa_py/serial.py
+++ b/pyvisa_py/serial.py
@@ -182,25 +182,23 @@ class SerialSession(Session):
         end_out, _ = self.get_attribute(ResourceAttribute.asrl_end_out)
         send_end, _ = self.get_attribute(ResourceAttribute.send_end_enabled)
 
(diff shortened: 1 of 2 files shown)


greyltc (commented):
Looks great! I think it should work. I'm just happy to see messages not dribbling out one byte at a time under any circumstance now ;-)

I still don't really get why there are two layers of termination character management (one here and one in https://github.com/pyvisa/pyvisa/blob/master/pyvisa/resources/messagebased.py ) but I guess that's okay if I don't understand the need for that.

I'll test this on my hardware shortly.

pyvisa_py/serial.py
```diff
@@ -182,25 +182,23 @@ def write(self, data: bytes) -> Tuple[int, StatusCode]:
         end_out, _ = self.get_attribute(ResourceAttribute.asrl_end_out)
         send_end, _ = self.get_attribute(ResourceAttribute.send_end_enabled)
 
-        try:
-            if end_out in (SerialTermination.none, SerialTermination.termination_break):
```
greyltc, line 186: @MatthieuDartiailh Why did you remove the check for `end_out` matching `SerialTermination.none` here? This means I fall into the `else` later and so my writes fail now with `Unknown value for VI_ATTR_ASRL_END_OUT: SerialTermination.none`

greyltc (changes requested):

pyvisa_py/serial.py
MatthieuDartiailh: Good catch. I was trying to avoid an empty branch but messed it up. I will fix it later today. Regarding the termchar the story is that for serial instrument sending the termchar can be part of the communication protocol itself while in othrr cases it is only required for the message to be meaningful. We could try avoiding duplicating it but I am unsure of the consequences.

MatthieuDartiailh:
Should be ready to go. The CIs issue is just a Codecov upload.

greyltc:
Confirmed working! Thanks!

greyltc (approved):

MatthieuDartiailh:
Bors r+

PR #287, opened 2020-11-19
protocols: improve timeout handling for vxi11

MatthieuDartiailh:
Based on last comments in https://github.com/pyvisa/pyvisa/issues/367

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: Continuous Integration
 on:
   schedule:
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
This does not seem to be the right solution. Testing against a Keysight system I discovered that I did get the expected answer but later than expected so either the timeout is improperly set for those systems or something or we need a tolerance around how long we wait for an answer. I won't have time to look into it soon but I will definitively come back to this.

PR #292, opened 2021-01-06
Fixing TCP/IP droping in from Dockercontainers

### LINKED ISSUE #285: RCP sessions remotely close after 5 minutes
When not using the device for 5 Minutes it closes the connection resulting in a `BrokenPipeError: [Errno 32] Broken pipe` when trying to send queries. I run this code in a Linux Docker container on a Windows host using the WSL2 backend.

When constantly using the connection this problem does not appear. Only after idling for 5 Minutes.

To Reproduce
--------------
```python
import pyvisa as visa
import time

rm = visa.ResourceManager('@py')
raw_inst = rm.open_resource('TCPIP0::10.0.1.21::inst0::INSTR')

raw_inst.query(':OUTP:STAT?')
time.sleep(60*6)
raw_inst.query(':OUTP:STAT?')
```

I already tried to set the socket to `keep_alive` by inserting `self.sock.setsockopt( socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)` into `pyvisa-py.protocols.rcp.RawTCPClient` inside the connect method with no success.

**Output of `pyvisa-info`**
```
/usr/local/lib/python3.8/site-packages/gpib_ctypes/gpib/gpib.py:54: UserWarning: GPIB library not found. Please manually load it using _load_lib(filename). All GPIB functions will raise OSError until the library is manually loaded.
  warnings.warn(message)
Machine Details:
   Platform ID:    Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.2.5
   Processor:      

Python:
   Implementation: CPython
   Executable:     /usr/local/bin/python
   Version:        3.8.6
   Compiler:       GCC 8.3.0
   Bits:           64bit
   Build:          Oct 13 2020 20:37:26 (#default)
   Unicode:        UCS4

PyVISA Version: 1.11.1

Backends:
   ivi:
      Version: 1.11.1 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.5.1
      USB INSTR: Available via PyUSB (1.1.0). Backend: N/A
      USB RAW: Available via PyUSB (1.1.0). Backend: N/A
      TCPIP INSTR: Available 
      TCPIP SOCKET: Available
```

Handfeger:
This fixes #285

I ran the tests and believe that other installations should not be affected (but not sure about windows). I am not sure if it is possible to write a test for this as you need to wait 5 minutes inside a docker container to experience this issue. I am happy to provide a test if anybody has an idea how it could be tested.

Another option to make this a option that can be accessed via highlevel function. I have found no way to access the sockets from outside. Maybe another option would be to expose the socket somehow to highlevel access points.

`isort` complaines about 2 files I did not touch: `/docs/source/conf.py` and `/pyvisa_py/testsuite/keysight_assisted_tests/test_tcpip_resources.py`. I did not fix this to not pollute this pull request.

- [x] Closes #285


MatthieuDartiailh:
The VISA specification mentions that VI_ATTR_TCPIP_KEEPALIVE should be supported by TCPIP::SOCKET and HiSLIP resources (a special case of INSTR), when enabled keep alive packets are used. However nothing is said for VXI-11 (the conventional TCPIP::INSTR). 

Currently PyVISA-py has no support for HiSLIP so let's not worry about that. For SOCKET it would be straightforward to port your solution to using the proper attribute. For INSTR, it may be worth providing a global flag in pyvisa-py to allow to opt-in that feature. How does that sound ? 

Let me know if you need any guidance.

pyvisa_py/protocols/rpc.py
```diff
@@ -476,6 +476,10 @@ def connect(self, timeout=5.0):
             "RawTCPClient: connecting to socket at (%s, %s)", self.host, self.port
         )
         self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
```
MatthieuDartiailh, line 479: Having some comments would be nice here

MatthieuDartiailh:
ping @Handfeger

Handfeger:
Sorry I'm currently learning for exams, so do not expect an answer before the middle of the month.

MatthieuDartiailh:
No problem.

Handfeger:
> Currently PyVISA-py has no support for HiSLIP so let's not worry about that. For SOCKET it would be straightforward to port your solution to using the proper attribute. For INSTR, it may be worth providing a global flag in pyvisa-py to allow to opt-in that feature. How does that sound ?
> 
> Let me know if you need any guidance.

As our instruments do not open a port for socket connection I fear, that the opt-in via global flag is our only option. Where would you define this global option? You mean setting a variable in the global scope and accessing that with pyvisa-py? That seems a bit hacky to me. Or would you set a global option in the session?

MatthieuDartiailh:
It would not be perfect but, if we want to avoid global variable at the module level (that could live under tcpip.py should we go that route nonetheless in the end), we could have that parameter lives on the Library object (defined in highlevel.py). However the library is never passed to the session so we need to either change that (in Session.__init__), or do the work in the library object which feels a bit weird since it is very specific to a session.

So I suggest:
-  having a new attribute on the VisaLibrary object `vxi11_use_keepalive`
- passing the library to session in their __init__ (be careful to alter all of them)
- apply your fix depending on the value of `vxi11_use_keepalive`, and store the attribute value on the session in case people need to inspect it later on
- implement the proper solution for SOCKET (ie your fix but based on VI_ATTR_TCPIP_KEEPALIVE  rather than `vxi11_use_keepalive`)

How does that sound ?

Handfeger:
Sounds reasonable. I'll try implementing that. You'll hear back from me if I run into any problems.

Handfeger:
Looking into the handling of attributes on the highlevel API I'm wondering if the introduction of a custom attribute would also work. This would mean to add checking for the custom attributes in `set_attribute` and `get_attribute`. But this way we could access the session and with that the socket of the instrument we want to activate keepalive packets on.

MatthieuDartiailh:
This is also an option but will require to define a new constant, since it does not exist at the VISA level. If you prefer that approach it is fine by me.

pyvisa_py/tcpip.py
```diff
@@ -59,6 +59,9 @@ class TCPIPInstrSession(Session):
     # for a specific kind of resource
     parsed: rname.TCPIPInstr
 
+    # Setting if keepalive has been activated
+    keepalive: bool
+
```
Handfeger, line 64: I'm not sure if this is the right to set this or if attributes are handled internally?

pyvisa_py/sessions.py
```diff
@@ -678,7 +679,7 @@ def set_attribute(
         session.
 
         Does a few checks before and calls before dispatching to
-        `_gst_attribute`.
+        `_set_attribute`.
```
Handfeger, line 682: Fixed a Typo

MatthieuDartiailh: Thanks

pyvisa_py/attributes.py
```diff
@@ -0,0 +1,37 @@
...
+    After the system detects that a connection was dropped, VISA returns a lost
+    connection error code on subsequent I/O calls on the session. The time required
+    for the system to detect that the connection was dropped is dependent on the
+    system and is not settable.
+
+    """
+
+    resources = [
+        (constants.InterfaceType.tcpip, "SOCKET"),
+        (constants.InterfaceType.tcpip, "INSTR"),
```
Handfeger, line 26: This is the only change to the original `AttrVI_ATTR_TCPIP_KEEPALIVE` by `PyVisa`. The rest is copied over.

Handfeger:
The approach I took was to overwrite the `VI_ATTR_TCPIP_KEEPALIVE` attribute to also be available for tcpip sessions of the `INSTR` type. This is done in the new file `attributes.py` which is imported in `sessions.py`. Because of the architecture of `PyVisa` attributes, importing is sufficient for overwriting it. After that all I did was to implement the attribute on session level in `tcpip.py`.

Are you okay with these decisions?

Below you see the test from inside the docker container. Code Test

pyvisa_py/tcpip.py
```diff
@@ -277,6 +281,35 @@ def _set_attribute(
             Return value of the library call.
 
         """
+
+        # In case of an environment with idle socket garbage collection (like docker)
+        # sockets need to be kept alive. Set <TODO: Global variable> to enable
```
MatthieuDartiailh, line 286: Comment need to be updated.

pyvisa_py/tcpip.py
MatthieuDartiailh: It looks fine to me.

MatthieuDartiailh (commented):
Overall it looks fine. I would love to see some addition to the documentation in particular since this allows pyvisa-py to deviate from VISA.
Out of curiosity would you have the bandwidth to test using Keysight IO Library suite (which is available for linux) in your docker image to see if the connection drops ? I am curious to know if they enabled the keepalive by default since it is not discussed as part of the vxi11 standard. If they do we should do the same.

MatthieuDartiailh:
bors try

Handfeger:
I'm happy to add documentation but wanted to check back if the code is fine first.

I also have a few further questions to get this pull request ready:
- Should I add an FAQ about broken pipes or is there some troubleshooting guide I missed?
- Where do you believe more documentation is needed?
- Do you think I should write a test for this feature? How would I go about that?
- flake8 is complaining about importing attributes without using it, but the import is needed so `__init_subclass__()` is executed once on this attribute to overwrite it. How do I handle this?

MatthieuDartiailh:
Regarding the documentation, you could add en entry under the FAQ of pyvisa-py. It is not ideal but since there is already en entry discussing use inside a VM it would work. We could also have an entry under Pyvisa documentation as another PR but mentionning the fact that the VXI11 part is pyvisa-py specific.

Regarding flake8 you can add `# noqa: ` followed by the error code to ignore.

Once this is done I will give bors another try to exercise your test.

Handfeger:
> Out of curiosity would you have the bandwidth to test using Keysight IO Library suite (which is available for linux) in your docker image to see if the connection drops ? I am curious to know if they enabled the keepalive by default since it is not discussed as part of the vxi11 standard. If they do we should do the same.

I am pretty sure this is not the standard option as we tried many different Visa systems and always ran into this particular problem. But to be honest I don't believe we tried the Keysight lib. So I might do that on the weekend.

I finished the documentation now so feel free to test run it all.

pyvisa_py/tcpip.py
```diff
@@ -277,6 +281,35 @@ def _set_attribute(
             Return value of the library call.
 
         """
+
+        # In case of an environment with idle socket garbage collection (like docker)
+        # sockets need to be kept alive. Set pyvisa.constants.ResourceAttribute.tcpip_keepalive to enable
+        # keepalive packets even for VXI11 protocol. To read more on this issue
+        # https://tech.xing.com/a-reason-for-unexplained-connection-timeouts-on-kubernetes-docker-abd041cf7e02
+        if attribute == constants.VI_ATTR_TCPIP_KEEPALIVE:
```
MatthieuDartiailh, line 289: Comparing to what we have in the SOCKET part, you appear to be setting many more socket attributes, would it make sense to mirror those for SOCKET resources ?

pyvisa_py/attributes.py
```diff
@@ -0,0 +1,37 @@
...
+
+:copyright: 2014-2020 by PyVISA-py Authors, see AUTHORS for more details.
+:license: MIT, see LICENSE for more details.
+"""
+
+from pyvisa import constants
+from pyvisa.attributes import BooleanAttribute
+
+
+class AttrVI_ATTR_TCPIP_KEEPALIVE(BooleanAttribute):
```
MatthieuDartiailh, line 14: I noticed you previously experimented with different solutions. Was there no way to rely on inheritance to get the values set in pyvisa ? Alternatively you could copy them over from pyvisa to avoid possible divergence in the future.

MatthieuDartiailh (commented):
Couple of small questions. I will start the run on the keysight bot.

MatthieuDartiailh:
bors try

MatthieuDartiailh:
Looks like the bot is down... I will look at it tomorrow.

pyvisa_py/tcpip.py
Handfeger: I believe that most of these options are needed for keepalive to fully function but I'm still not sure as I've not understood these keepalive packets fully. As I understand it, the further options tell the socket to start after one minute of being idle to send keepalive packets every minute. The first option was crucial for us as the standard was way more.

The last option is the number of retries and not really needed i guess.

I can copy them to the socket as well.

pyvisa_py/attributes.py
Handfeger: To be honest I tried a lot and now that I am thinking about it I'm not sure why I'm not inheriting from the attribute directly. I will try that out as it would cut down the lines of code significantly. An changes from pyvisa would be handed down.

Handfeger:
Inheriting is in and everything still works.

Screenshot 2021-04-28 at 13 22 42

pyvisa_py/tcpip.py
```diff
@@ -692,6 +725,9 @@ def _set_tcpip_keepalive(
             self.interface.setsockopt(
                 socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1 if attribute_state else 0
             )
+            self.interface.sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 60)
```
MatthieuDartiailh, line 728: Since the setting is of the other options is not conditional on the value (I assume this is fine), it may be worth doing it earlier (and the same would apply for the INSTR too).

Handfeger: I'm not sure if it is necessary as this only applies if you enable keepalive packets. I'm also pretty positive, that this function would only be called once in the lifetime of a session.

MatthieuDartiailh: I guess you are right in practice people are unlikely to enable and disable this many times.

MatthieuDartiailh:
Not sure you can see it but here is the log of the keysight bot https://dev.azure.com/pyvisa/pyvisa-py/_build/results?buildId=192&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=a1c457be-0fbe-5525-b6b0-ac3cf58f8d6c&l=135. 

Basically your test do not pass, because `self.instr` is not the pyvisa-py internal session object but the normal pyvisa resource object. You will need something like `self.instr.visalib.sessions[self.instr.session]` to obtain the object on which `keepalive` exists.

Handfeger:
Is there any way to test my function locally? `pytest` doesn't seem to test this particular function for me and I don't want to waste your time with failing code.

MatthieuDartiailh:
In order for those test to run you need to set an env var (see the \_\_init\_\_.py of the keysight assisted folder of pyvisa). If I spell it from memory I will get it wrong. You will also see the test IP.

MatthieuDartiailh:
bors try

MatthieuDartiailh:
I happen to have some serious issue on the bot at the moment and may need to swap out a computer so I do not know I will be able to have it back online. If you can get the test to pass locally, great, otherwise we can merge as is and open a new issue about the test and I will look into it when the bot is back online.

Handfeger:
We are currently running an experiment on the hardware so the earliest point would be end of the week, so maybe we open a new issue for the test.

MatthieuDartiailh:
bors try

MatthieuDartiailh:
Apparently windows believes it managed to fix a disk issue on the bot, so I will try to run the bot one last time and if it fails for something unrelated to this I will merge and open an issue.

MatthieuDartiailh:
The build failed but that does not look related to your work. I will try to have a look nonetheless but I will do my best to merge shortly.

Handfeger:
We currently have a workaround so there is not that much time pressure on this issue.

MatthieuDartiailh:
Since I have had no time to look into the bot issue I will merge as is. Thanks for your work @Handfeger an sorry for the delay.

PR #297, opened 2021-02-13
tcpip: List resources to fix #165

### LINKED ISSUE #165: 0.3.1 does not discover TCPIP INST
We used to communicate with an ASTERION AST-501 power supply over TCPIP using the default "@ni" backend. Since there seems to be an issue with longer command strings transmitted over TCPIP, we wanted to test the pyVIsa-py backend but had to notice that our instrument is not discovered by the resource_manager.

Is this a bug? Is it related to the device not being compatible with pyVisa-py? Or do we need to configure something else to make it work? Any suggestions on how we get closer to the root cause? 

The ni-backend shows the device as 'TCPIP0::169.254.180.2::inst0::INSTR'.

pyvisa-info
```
Machine Details:
   Platform ID:    Windows-7-6.1.7601-SP1
   Processor:      x86 Family 6 Model 42 Stepping 7, GenuineIntel

Python:
   Implementation: CPython
   Executable:     c:\users\[...]\python36-32\python.exe
   Version:        3.6.0
   Compiler:       MSC v.1900 32 bit (Intel)
   Bits:           32bit
   Build:          Dec 23 2016 07:18:10 (#v3.6.0:41df79263a11)
   Unicode:        UCS4

PyVISA Version: 1.9.1

Backends:
   ni:
      Version: 1.9.1 (bundled with PyVISA)
      #1: c:\windows\system32\visa32.dll:
         found by: auto
         bitness: 32
         Vendor: National Instruments
         Impl. Version: 18874368
         Spec. Version: 5244928
   py:
      Version: 0.3.1
      USB INSTR:
         No module named 'usb'
      USB RAW:
         No module named 'usb'
      TCPIP INSTR: Available
      TCPIP SOCKET: Available
      GPIB INSTR:
         Please install linux-gpib to use this resource type.
         No module named 'gpib'
```
pip list
```
Package        Version
-------------- ---------
altgraph       0.15
beautifulsoup4 4.6.0
comtypes       1.1.4
future         0.16.0
macholib       1.9
pefile         2017.11.5
pip            10.0.1
PyInstaller    3.3.1
pypiwin32      220
PyQt5          5.10.1
pyqt5-tools    5.9.0.1.2
pyserial       3.4
PyVISA         1.9.1
PyVISA-py      0.3.1
pywifi         1.1.10
setuptools     28.8.0
sip            4.19.8
tabulate       0.8.2
```

ekigwana:
Got the snippet from https://github.com/python-ivi/python-vxi11/blob/master/vxi11/vxi11.py#L513

Signed-off-by: Edward Kigwana <ekigwana@gmail.com>

- [x] Closes #165

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,11 @@
 PyVISA-py Changelog
 ===================
 
+0.5.3 (unreleased)
+------------------
+
+- List TCPIP resources to fix #165
+
 0.5.2 (04-02-2020)
 ------------------
 
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -63,15 +63,19 @@ class TCPIPInstrSession(Session):
     def list_resources() -> List[str]:
         pmap = rpc.BroadcastUDPPortMapperClient("255.255.255.255")
         pmap.set_timeout(1)
-        resp = pmap.get_port(
-            (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
-        )
-        res = [r[1][0] for r in resp if r[0] > 0]
-        res = sorted(res, key=lambda ip: tuple(int(part) for part in ip.split(".")))
-        # TODO: Detect GPIB over TCPIP
-        res = ["TCPIP::{}::INSTR".format(host) for host in res]
 
-        return res
+        try:
+            resp = pmap.get_port(
+                (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
+            )
+            res = [r[1][0] for r in resp if r[0] > 0]
+            res = sorted(res, key=lambda ip: tuple(int(part) for part in ip.split(".")))
+            # TODO: Detect GPIB over TCPIP
+            res = ["TCPIP::{}::INSTR".format(host) for host in res]
+
+            return res
+        except rpc.RPCError:
+            return []
 
     def after_parsing(self) -> None:
         # TODO: board_number not handled
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -61,8 +61,17 @@ class TCPIPInstrSession(Session):
 
     @staticmethod
     def list_resources() -> List[str]:
-        # TODO: is there a way to get this?
-        return []
+        pmap = rpc.BroadcastUDPPortMapperClient("255.255.255.255")
+        pmap.set_timeout(1)
+        resp = pmap.get_port(
+            (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
+        )
+        res = [r[1][0] for r in resp if r[0] > 0]
+        res = sorted(res, key=lambda ip: tuple(int(part) for part in ip.split(".")))
+        # TODO: Detect GPIB over TCPIP
+        res = ["TCPIP::{}::INSTR".format(host) for host in res]
+
+        return res
 
     def after_parsing(self) -> None:
         # TODO: board_number not handled
--- a/pyvisa_py/testsuite/keysight_assisted_tests/test_resource_manager.py
+++ b/pyvisa_py/testsuite/keysight_assisted_tests/test_resource_manager.py
@@ -3,7 +3,12 @@
 
 """
 import pytest
-from pyvisa.testsuite.keysight_assisted_tests import copy_func, require_virtual_instr
+from pyvisa.rname import ResourceName
+from pyvisa.testsuite.keysight_assisted_tests import (
+    RESOURCE_ADDRESSES,
+    copy_func,
+    require_virtual_instr,
+)
 from pyvisa.testsuite.keysight_assisted_tests.test_resource_manager import (
     TestResourceManager as BaseTestResourceManager,
     TestResourceParsing as BaseTestResourceParsing,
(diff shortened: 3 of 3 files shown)


MatthieuDartiailh:
Thanks for your contributions. Did you get a chance to test it?

I hope to bring the keysight build bot back online in the next two weeks. I will likely wait for it before merging your work. Do not hesitate to ping me if I were to fall behind schedule.

ekigwana:
Yes, I tested it all using the PyVISA frontend and it works. I don't have any tcpip bridge adapters to old equipment so I could not code for that. The snippent I references addresses that by attempting to open connections. This solves a major irritation when you multiple calibration stations with the same equipment on a dhcp network and folks have to run around to figure out what the new address is. Now they only have to get the equipment serial information, that + IDN and voila no need to know IP address.

MatthieuDartiailh:
Thanks. I am slowly bringing the keysight based bot back online so I will try to add test for this PR and then give it a run. Do not hesitate to ping me if I take too long.

MatthieuDartiailh:
bors try

MatthieuDartiailh:
Interestingly the self-hosted buildbot I have has 2 network card and your PR is picking up the resources on only a single one of them, could be worth investigating since that is a usual configuration ? Also It would be best to return normalized resource name using `pyvisa.rname.to_canonical_name` rather than names that do not include the port and the lan resource name.

MatthieuDartiailh:
Playing with this locally and comparing to the output of NI-MAX, I also observe discrepancies. I am testing on a Windows 10 machine that acts as an instrument itself. NI finds the instrument under the IP 127.0.0.1 and 192.168.1.22, on the other hand pyvisa-py finds under 192.168.56.1. I am not sure what is the issue here but I will try to run some more tests during the week.

eosti:
I did a little testing with this PR and found that it will only broadcast on one interface. That might explain the oddities with multiple network cards and potentially with the local device not showing up as expected (broadcasts can't respond to themselves). 

I was testing with a power supply on the LAN over WiFi and a scope connected directly with a USBC->Ethernet adapter. 

- Both interfaces connected: only the PSU showed up
- Ethernet adapter disconnected: only PSU showed up, still
- Wifi disabled: only scope shows up

I assume that this is because WiFi was the default interface when it was on, so broadcasts would go through there. When WiFi is disabled, the Ethernet adapter becomes the default interface, so now broadcasts will go through that. When both are on, broadcasts _only_ go through the default interface. 

I'm not familiar with the RPC protocol library; is there a way to specify what interface to use for broadcast packets? That way you could get a list of all valid interfaces and then iterate through them, or something like that. Or maybe there's a better way, I'm really not an expert in broadcasts.

MatthieuDartiailh:
The rpc library is basically bundled and has been already customized so I would not worry too much about it. Looking into it, it appears to simply call stdlib socket but I see nothing in there specific to adapters. It may be worth looking directly at the socket documentation.

eosti:
Based on a bit of research, there are two ways to go about specifying an interface for broadcast.

1. Force a specific interface by using `sock.setsockopt(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, 'iface')`
2. Exploit the OS's routing table by broadcasting to the each subnet's broadcast address, instead of the global `INADDR_BROADCAST` (255.255.255.255). This will automatically use the correct interface. 

Method 1 needs sudo access, so that rules it out pretty quickly for me. Method 2 is doable, but it needs both the IP address and the subnet mask of each interface. IP address can be obtained from `socket`, but I haven't found a native way to get it otherwise. 

Using the library `netifaces`, I was able to generate a list of broadcast addresses, and then iterate through them with    `pmap = rpc.BroadcastUDPPortMapperClient(ip)`. And I was able to see all devices across all subnets! So it _is_ possible. My only hesitation is that `netifaces` isn't a standard library, so it'd need to be installed as a dependency. I don't know what the etiquette is when it comes to adding dependencies, so let me know what the best move here would be. 

For reference, the snippet I've been using to generate the list of broadcast addresses:

```
import netifaces 
import ipaddress

broadcast_addr = []

for interface in netifaces.interfaces():
    for link in netifaces.ifaddresses(interface).get(netifaces.AF_INET, ()):
        addr = link['addr']
        mask = link['netmask']
        network = ipaddress.IPv4Network(addr + '/' + mask, strict=False)
        broadcast_addr.append(network.broadcast_address)

print(broadcast_addr)
```

You can then loop the discovery process for each broadcast address. I'm happy to make a PR (or edit this one), with the caveats that it would require the use of `netifaces` and that it'd need a bit of cleaning up (Python isn't my strongest language).

MatthieuDartiailh:
I am not opposed to dependencies in general but netifaces is not maintained anymore (the repo has been archived) and does not provide wheels for Python 3.10.

It appears psutil exposes similar albeit more limited capabilities. Would it be possible to use it instead ?

eosti:
Oh, good point. Got it to work with psutil:

```
import psutil
import socket
import ipaddress

broadcast_addr = []

for interface, snics in psutil.net_if_addrs().items():
    for snic in snics:
        # Restrict to IPv4
        if snic.family is socket.AF_INET:
            addr = snic.address
            mask = snic.netmask
            network = ipaddress.IPv4Network(addr + '/' + mask, strict=False)
            broadcast_addr.append(network.broadcast_address)

print(broadcast_addr)
```

I've noticed that broadcasting and waiting for responses takes a bit of time, and scales linearly with the number of IP addresses you have. It might make sense to make the default behaviour to broadcast on the default interface only (e.g. 255.255.255.255) and then add a switch to broadcast on all interfaces. This way, list_resources only is slowed down if you know that your specific setup requires it.

MatthieuDartiailh:
Sounds good.

Could you either update this PR or make a new one ?

eosti:
I'm working on a PR right now. What would the best way be to pass a boolean into `list_resources` to determine if a full scan should be run or not? Ideally it'd be something that could be set using pyvisa. 

Is there a way to access a ResourceManager's attributes from pyvisa? Could you make a class variable `full_scan`, then set it using `rm.[something].TCPIPInstrSession.full_scan = True`? Or is this totally the wrong approach?

PR #315, opened 2022-01-11
fix(logging): Fix logging error and add missing argument

lovetheguitar:
When depending on the logging class, to format the logging message, arguments need to get passed separately, unlike with direct string formatting where a tuple with all arguments is expected.


MatthieuDartiailh:
Thanks for your patch ! 

Can you run black to appease the CIs and then I will merge ?

lovetheguitar:
Very welcome. Thanks for the good work.
Formatting done ✔️ Sorry for not running it before, I could not imagine it would reformat something with so little change. 😄

PR #322, opened 2022-04-15
VXI11 bug fixes

bobmcnamara:
- [ n/a] Closes # (insert issue number if relevant)
- [ X] Executed ``black . && isort -c . && flake8`` with no errors
- [ n/a] The change is fully covered by automated unit tests
- [ n/a] Documented in docs/ as appropriate
- [ X] Added an entry to the CHANGES file

### DIFF
--- a/pyvisa_py/protocols/rpc.py
+++ b/pyvisa_py/protocols/rpc.py
@@ -279,6 +279,17 @@ class Client(object):
 # Record-Marking standard support
 
 
+def _sendto(sock, data, address):
+    """
+    loops calling sock.sendto() until all data is sent.
+    """
+    ptr = 0
+    while data[ptr:]:
+        ptr += sock.sendto(data[ptr:], address)
+
+
+# XXX sendfrag() should have been deleted when it was inlined into
+# _sendrecord() during refactoring
 def sendfrag(sock, last, frag):
     x = len(frag)
     if last:
@@ -292,7 +303,7 @@ def _sendrecord(sock, record, fragsize=None, timeout=None):
     if timeout is not None:
         r, w, x = select.select([], [sock], [], timeout)
         if sock not in w:
-            msg = "socket.timeout: The instrument seems to have stopped " "responding."
+            msg = "socket.timeout: The instrument seems to have stopped responding."
             raise socket.timeout(msg)
 
     last = False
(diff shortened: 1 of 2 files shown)


pyvisa_py/protocols/rpc.py
```diff
@@ -569,14 +579,19 @@ def set_reply_handler(self, reply_handler):
     def set_timeout(self, timeout):
         self.timeout = timeout  # Use None for infinite timeout
 
+    def sock_sendto(self, data, address):
+        ptr = 0
+        while len(data[ptr:]) > 0:
```
MatthieuDartiailh, line 584: You could simply write

```suggestion
        while data[ptr:]:
```

pyvisa_py/protocols/rpc.py
```diff
@@ -979,8 +994,13 @@ def loop(self):
         while 1:
             self.session()
 
+    def sock_sendto(self, data, address):
```
MatthieuDartiailh, line 997: I would prefer to avoid duplicating this. Could you refactor this as a function taking the socket as first argument ?

MatthieuDartiailh (changes requested):
I left minor comments but it looks great. Please also move the changelog entry under 0.5.3.

I will do my best to fix the formatting issues on main in the meantime.

pyvisa_py/protocols/rpc.py
bobmcnamara: I agree, it's cleaner that way.  I'd feel more comfortable if I had a way to test the code that uses sock.sendto(), but I don't even see a way for the UDP code to get executed.  Maybe it should just be deleted?

I recall running into some device in the past that supported UDP connections, but I don't recall which instrument it was.  Most instruments I've worked with only support TCPIP connections.

MatthieuDartiailh: When pyvisa-py was started (before I got involved), the rpc file was pulled in as a basic implementation of RPC upon which VXI-11 is based. But it is true that large swabs of it are absolutely useless since VXI -11 does not requires a complete RPC implementation. However due to the little time I have to work on this and the difficulty to test things to avoid regression I never deleted the useless bits of that file.

MatthieuDartiailh:
I pushed a formatting fix on main so I will merge. Thanks again for the patch !

PR #326, opened 2022-06-10
Add TCPIP instrument discovery

### LINKED ISSUE #165: 0.3.1 does not discover TCPIP INST

eosti:
Continues the work done in #297 by broadcasting on all interfaces. Tweaks the RPC library to add methods for only sending and only receiving broadcasts, so that the timeout period can be shared across all interfaces. 

Adds optional dependency `psutil`, which is needed to discover instruments on all interfaces. Otherwise, the only instruments on the default interface will be discovered. 

- [x] Closes #165, fixes #297

### DIFF
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -15,7 +15,7 @@ from typing import Any, List, Optional, Tuple
 
 # Let psutil be optional dependency
 try:
-    import psutil
+    import psutil  # type: ignore
 except ImportError:
     psutil = None
 
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -85,22 +85,28 @@ class TCPIPInstrSession(Session):
             # If psutil unavailable fallback to default interface
             broadcast_addr.append("255.255.255.255")
 
-        try:
-            pmap_list = [rpc.BroadcastUDPPortMapperClient(ip) for ip in broadcast_addr]
-            for pmap in pmap_list:
-                pmap.set_timeout(0)
+        pmap_list = [rpc.BroadcastUDPPortMapperClient(ip) for ip in broadcast_addr]
+        for pmap in list(pmap_list):
+            pmap.set_timeout(0)
+            try:
                 pmap.send_port(
                     (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
                 )
+            except rpc.RPCError:
+                pmap_list.remove(pmap)
 
-            # Timeout for responses
-            time.sleep(1)
+        # Timeout for responses
+        time.sleep(1)
 
-            all_res = []
-            for pmap in pmap_list:
+        all_res = []
+        for pmap in pmap_list:
+            try:
                 resp = pmap.recv_port(
                     (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
                 )
+            except rpc.RPCError:
+                pass
+            else:
                 res = [r[1][0] for r in resp if r[0] > 0]
                 res = sorted(
                     res, key=lambda ip: tuple(int(part) for part in ip.split("."))
@@ -108,8 +114,6 @@ class TCPIPInstrSession(Session):
                 # TODO: Detect GPIB over TCPIP
                 res = ["TCPIP::{}::INSTR".format(host) for host in res]
                 all_res.extend(res)
-        except rpc.RPCError:
-            return []
 
         return all_res
 
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -6,17 +6,17 @@
 :license: MIT, see LICENSE for more details.
 
 """
+import ipaddress
 import random
 import select
 import socket
-import ipaddress
 import time
 from typing import Any, List, Optional, Tuple
 
 # Let psutil be optional dependency
 try:
     import psutil
-except:
+except ImportError:
     psutil = None
 
 from pyvisa import attributes, constants, errors, rname
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -9,8 +9,15 @@
 import random
 import select
 import socket
+import ipaddress
 import time
 from typing import Any, List, Optional, Tuple
+
+# Let psutil be optional dependency
+try:
+    import psutil
+except:
+    psutil = None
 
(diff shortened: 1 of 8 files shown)


MatthieuDartiailh:
I will do my best to review and test with the keysight tools hopefully Monday.

MatthieuDartiailh:
Sorry I got no time to look at this last week. I will do it ASAP but I am quite busy ATM.

pyvisa_py/tcpip.py
```diff
@@ -64,8 +71,48 @@ class TCPIPInstrSession(Session):
...
+            for interface, snics in psutil.net_if_addrs().items():
+                for snic in snics:
+                    if snic.family is socket.AF_INET:
+                        addr = snic.address
+                        mask = snic.netmask
+                        network = ipaddress.IPv4Network(addr + "/" + mask, strict=False)
+                        broadcast_addr.append(str(network.broadcast_address))
+        else:
+            msg = "To discover devices on all interfaces, please install psutil"
+            Session.register_unavailable(constants.InterfaceType.tcpip, "INSTR", msg)
```
MatthieuDartiailh, line 86: No doing this make TCPIP resources unavailable without psutil !

pyvisa_py/tcpip.py
```diff
@@ -64,8 +71,48 @@ class TCPIPInstrSession(Session):
...
+                    (vxi11.DEVICE_CORE_PROG, vxi11.DEVICE_CORE_VERS, rpc.IPPROTO_TCP, 0)
+                )
+                res = [r[1][0] for r in resp if r[0] > 0]
+                res = sorted(
+                    res, key=lambda ip: tuple(int(part) for part in ip.split("."))
+                )
+                # TODO: Detect GPIB over TCPIP
+                res = ["TCPIP::{}::INSTR".format(host) for host in res]
+                all_res.extend(res)
+        except rpc.RPCError:
```
MatthieuDartiailh, line 112: On Windows I get a crash with an os error (OSError: [WinError 10051] A socket operation was attempted to an unreachable network). Need to be fixed either in rpc.py or here. rpc.py may make more sense

pyvisa_py/testsuite/keysight_assisted_tests/test_resource_manager.py
```diff
@@ -14,9 +19,15 @@
...
-    )
+    def test_list_resource(self):
+        """Test listing the available resources.
+        The bot supports only TCPIP and of those resources we expect to be able
+        to list only INSTR resources not SOCKET.
+        """
+        # Default settings
+        resources = self.rm.list_resources()
+        for v in (v for v in RESOURCE_ADDRESSES.values() if v.endswith("INSTR")):
+            assert str(ResourceName.from_string(v)) in resources
```
MatthieuDartiailh, line 30: Surprisingly for an instrument on the loop back I get the default gateway 192.168.1.56 which is annoying but I have no idea how to fix it.

MatthieuDartiailh (changes requested):
I tested locally and noticed a couple issues. I will have a second look once you address them and soryy for the slow response.

pyvisa_py/tcpip.py
eosti: Got it, wasn't sure if I should have added this or not. The default behaviour if `psutil` is not found is now to fail silently, and only broadcast on the default interface.

pyvisa_py/testsuite/keysight_assisted_tests/test_resource_manager.py
eosti: Hmmm, I'll need to see if I can reproduce this on my side. It might be a limitation of UDP broadcasts -- a device cannot respond to its own broadcast (so if a machine has a virtual instrument hosted on the same IP as the discovery process, it can't see itself). I'll look into this.

MatthieuDartiailh: The result is not wrong (I can actually address the device through that IP address) just unexpected. But if you figure out how to fix it all the best.

MatthieuDartiailh:
Sorry for the very very long delay. I finally took some time to play with your code both with and without psutil installed. For my stupid fake instrument case I get many hits with psutil including 127.0.0.1 (which is the address I expect). Those addresses do not all match loopback as defined in the IPv4 standard. There are all valid so there is no meaningful way to filter them.

This is contrived case and the fact you code gives a slightly unexpected answer should not all back your work. I want to apologize again for the long delay and thank you for your contribution.

PR #331, opened 2022-07-27
addition of HiSLIP specification V1.1 (current version is 2.0)

bobmcnamara:


MatthieuDartiailh:
Could you make a separate PR for VICP ? I will review this PR before the end of the week but I would prefer to focus on a single thing at a time.

pyvisa_py/protocols/rpc.py
```diff
@@ -454,7 +454,8 @@ class RawTCPClient(Client):
 
     def __init__(self, host, prog, vers, port, open_timeout=5000):
         Client.__init__(self, host, prog, vers, port)
-        self.connect((open_timeout / 1000.0) + 1.0)
+        open_timeout = open_timeout or 5000
```
MatthieuDartiailh, line 457: Avoid turning a 0 into 5000
```suggestion
        open_timeout = open_timeout if open_timeout is not None else 5000
```

pyvisa_py/sessions.py
```diff
@@ -122,8 +122,8 @@ def close(self) -> StatusCode:
     #: Session type as (Interface Type, Resource Class)
     session_type: Tuple[constants.InterfaceType, str]
 
-    #: Timeout in seconds to use when opening the resource.
-    open_timeout: Optional[float]
+    #: Timeout in milliseconds to use when opening the resource.
+    open_timeout: Optional[int]
```
MatthieuDartiailh, line 126: I will need to check how that interact with other backends (which are far harder to test too).

pyvisa_py/tcpip.py
```diff
@@ -41,6 +41,203 @@
...
+                resource_manager_session, resource_name, parsed, open_timeout
+            )
+        else:
+            obj = TCPIPInstrVxi11(
+                resource_manager_session, resource_name, parsed, open_timeout
+            )
+        return obj
+
+
+@Session.register(constants.InterfaceType.tcpip, "HISLIP")
```
MatthieuDartiailh, line 69: This does not really makes sense since HISLIP is not a recognized VISA resource class. I would simply remove it.

pyvisa_py/tcpip.py
```diff
@@ -41,6 +41,203 @@
...
+
+        if parsed.lan_device_name.lower().startswith("hislip"):
+            obj = TCPIPInstrHiSLIP(
+                resource_manager_session, resource_name, parsed, open_timeout
+            )
+        else:
+            obj = TCPIPInstrVxi11(
+                resource_manager_session, resource_name, parsed, open_timeout
+            )
+        return obj
```
MatthieuDartiailh, line 66: We could allow TCPIPInstrHiSLIP, TCPIPInstrVxi11 to subclass this class by modifying the method as such:
```suggestion
    def __new__(
        cls,
        resource_manager_session: VISARMSession,
        resource_name: str,
        parsed=None,
        open_timeout: Optional[int] = None,
    ):
        if cls is not TCPIPInstrSession:
            return super().__new__(cls, resource_manager_session, resource_name, parsed, open_timeout)

        if parsed is None:
            parsed = rname.parse_resource_name(resource_name)

        if parsed.lan_device_name.lower().startswith("hislip"):
            cls = TCPIPInstrHiSLIP
        else:
            cls = TCPIPInstrVxi11
            
        return cls(resource_manager_session, resource_name, parsed, open_timeout)
```

This would allow to share common parts more easily.

pyvisa_py/tcpip.py
```diff
@@ -41,6 +41,203 @@
...
+class TCPIPInstrHiSLIP(Session):
+    """A TCPIP Session built on socket standard library using HiSLIP protocol."""
+
+    # Override parsed to take into account the fact that this class is only used
+    # for a specific kind of resource
+    parsed: rname.TCPIPInstr
+
+    @staticmethod
+    def list_resources() -> List[str]:
+        # TODO: is there a way to get this?
```
MatthieuDartiailh, line 79: There is a pending PR for VXI-11, would something similar be envisioned for HiSLIP ?

pyvisa_py/tcpip.py
```diff
@@ -68,15 +265,18 @@ def list_resources() -> List[str]:
         return []
 
     def after_parsing(self) -> None:
-        # TODO: board_number not handled
```
MatthieuDartiailh, line 71: This is still true and should be preserved.

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+HEADER_SIZE = struct.calcsize(HEADER_FORMAT)
+
+
+class Struct(dict):
+    def __init__(self, **kwargs):
+        super(Struct, self).__init__(**kwargs)
+        self.__dict__ = self
+
+
+class Instrument:  # pylint: disable=too-many-instance-attributes
```
MatthieuDartiailh, line 115: pylint is not run on this repo so I would prefer to avoid comment specific to it.

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+        self._async = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self._async.connect((ip_addr, port))
+        self._async.settimeout(timeout)
+        self._async.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+        self._async_init = self.async_initialize(session_id=init.session_id)
+
+        # get the maximum message size
+        max_msg_size = self.async_maximum_message_size(1 << 24)
+        # print("HEADER_SIZE = %s, async_maximum_message_size = %s" % (HEADER_SIZE, max_msg_size))
+        # print("max_payload_size = %s" % (max_msg_size - HEADER_SIZE))
```
MatthieuDartiailh, line 149: outdated ? if yes please remove them

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+    @timeout.setter
+    def timeout(self, val):
+        """sets the timeout value in seconds for both the sync and async sockets"""
+        self._timeout = val
+        self._sync.settimeout(self._timeout)
+        self._async.settimeout(self._timeout)
+
+    def send(self, data):
+        """Sends the data on the synchronous channel.  More than one packet
+        may be necessary in order to not exceed max_payload_size."""
```
MatthieuDartiailh, line 179: I usually format docstring following the Numpy format so I am suggesting changes to keep the style homogeneous.
```suggestion
        """Sends the data on the synchronous channel.  
        
        More than one packet may be necessary in order to not exceed 
        max_payload_size.
        
        """
```

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+    # MEMBER FUNCTIONS #
+    # ================ #
+
+    def close(self):
+        self._sync.close()
+        self._async.close()
+
+    @property
+    def timeout(self):
+        """returns the timeout value in seconds for both the sync and async sockets"""
```
MatthieuDartiailh, line 167: A property is basically an attribute so I prefer the docstring to reflect this.
```suggestion
        """Timeout value in seconds for both the sync and async sockets"""
```

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+        return self._timeout
+
+    @timeout.setter
+    def timeout(self, val):
+        """sets the timeout value in seconds for both the sync and async sockets"""
```
MatthieuDartiailh, line 172:
```suggestion
        """Timeout value in seconds for both the sync and async sockets"""
```

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+            data_view = data_view[bytes_sent:]
+            num_bytes_to_send -= bytes_sent
+
+        return len(data)
+
+    def receive(self, max_len=4096):
+        """
+        receive data on the synchronous channel, terminating after
+        max_len bytes or after receiving a DataEnd message
+        """
```
MatthieuDartiailh, line 202:
```suggestion
        """Receive data on the synchronous channel.
        
        Terminate after max_len bytes or after receiving a DataEnd message.
        """
```

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+
+        # if we don't already have a Receiver object, create one
+        if self.receiver is None:
+            self.receiver = Receiver(self._sync, self.last_message_id)
```
MatthieuDartiailh, line 206: Is there an advantage to delaying the receiver creation in this fashion. Could it not be created as part of the initialization ?

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+        #     S->C: AsyncMaxMsgSizeResponse
+        header = struct.pack(
+            "!2sBBIQQ", b"HS", MESSAGETYPE["AsyncMaxMsgSize"], 0, 0, 8, size
+        )
+        self._async.sendall(header)
+        response_header = receive_header(
+            self._async, expected_message_type="AsyncMaxMsgSizeResponse"
+        )
+        # print("max_msg_size = %s, type(%s)" % (response_header.max_msg_size,
+        #                                       type(response_header.max_msg_size)))
```
MatthieuDartiailh, line 283: remove ?

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+        bytes_recvd += data_len
+        view = view[data_len:]
+
+    if bytes_recvd > recv_len:
+        raise MemoryError("socket.recv_into scribbled past end of recv_buffer")
+
+
+def receive_header(  # noqa: C901
+    sock, expected_message_type=None
+):  # pylint: disable=too-many-statements,too-many-branches
```
MatthieuDartiailh, line 520: Same comment about pylint

pyvisa_py/protocols/hislip.py
```diff
@@ -0,0 +1,714 @@
...
+import socket
+import struct
+import time
+from typing import Dict, Union
+
+PORT = 4880
+
+LookupTable = Dict[Union[int, str], Union[int, str]]
+
+MESSAGETYPE: LookupTable = {
```
MatthieuDartiailh, line 16: What advantage to do you see in merging the forward (key-> value) and backward (value->key) mapping. It looks to me like it is always clear which one should be used. We could keep them separate and get stricter typing.

MatthieuDartiailh (commented):
It looks generally really good. I left some small comments.

Regarding timeouts I will have to double check possible cross-talk with other sessions.

MatthieuDartiailh:
Also what is the empty setup file ?

bobmcnamara:
> Also what is the empty setup file ?

The contents of setup.py shouldn't have changed...I only changed the mode to make it executable.

pyvisa_py/sessions.py
bobmcnamara: From my examination of the code, this appears to have been a bug.  The open_timeout is passed in from VISA where it is always expressed as an int in milliseconds (see constants.VI_TMO_IMMEDIATE and constants.VI_TMO_INFINITE).

pyvisa_py/tcpip.py
bobmcnamara: Registering a non-existent VISA resource class may not be desirable, but without the Session.register decorator we won't get the session_type attribute defined, and that causes a problem in Session.set_attribute.  It looks like fixing that would require some major refactoring.

pyvisa_py/tcpip.py
bobmcnamara: If we make TCPIPInstrHiSLIP and TCPIPInstrVxi11 subclass TCPIPInstrSession instead of Session, we add a couple of complications for absolutely no benefit.

The first complication is having to test if cls is TCPIPInstrSession, then calling super().__new__().  (And that didn't work for me, I had to call Session.__new__() instead.)

The second complication is that the return value has to be cls.__new__() instead of cls().  Otherwise __init__ gets called twice on the resulting object. 

After trying it both ways, my preference is for TCPIPInstrHiSLIP and TCPIPInstrVxi11 to subclass Session.  It just feels like there's less dunder magic going on.  But it will work either way.

pyvisa_py/tcpip.py
bobmcnamara: Agree, deleting it was an unintentional error on my part.

pyvisa_py/protocols/hislip.py
bobmcnamara: For me the only advantage is that I don't have to think of two good, unique names.  Naming things is hard, and I'm lazy.  Think of it as a form of polymorphism.

pyvisa_py/protocols/hislip.py
bobmcnamara: This needs some work.  Let me take a look and play with a couple of ideas.

pyvisa_py/tcpip.py
bobmcnamara: It looks like everywhere I typed \_\_new\_\_ it got replaced with a bold new.

MatthieuDartiailh:
Out of curiosity why did you need to make setup.py executable in the first place ? With PEP517 one should always install through pip.

pyvisa_py/protocols/hislip.py
MatthieuDartiailh: Fair enough

pyvisa_py/sessions.py
MatthieuDartiailh: I think you are indeed right thanks for the fix.

pyvisa_py/tcpip.py
MatthieuDartiailh: The problem is that it will mess up the access to the resource class VISA attribute. At this point it sounds just better to manually add `cls.session_type = (interface_type, resource_class)` by hand rather than relying on the registering mechanism. It is not perfect but it is more correct from a user perspective.

pyvisa_py/tcpip.py
MatthieuDartiailh: My main concern was to avoid code duplication between VXI11 and HiSLIP since the interface object in both cases can have similar APIs but I may be over complicating things here. So let's just keep your current solution.

MatthieuDartiailh:
@bobmcnamara please ping me when this is ready for a new review.

bobmcnamara:
> Out of curiosity why did you need to make setup.py executable in the first place ? With PEP517 one should always install through pip.

Good question.  I was working on releasing the pyvicp package to PyPI (my first!) and I wasn't really understanding the process.  I started experimenting with the pyvisa-py package and must have made setup.py executable after I got tired of typing "python3 setup.py".  But I can't even recall why I was doing that in the first place.

bobmcnamara:
> @bobmcnamara please ping me when this is ready for a new review.

@MatthieuDartiailh when you get a chance, please take another look.

pyvisa_py/protocols/hislip.py
```diff
@@ -40,61 +38,71 @@
...
+    "enableRemote": 1,
+    "disableAndGTL": 2,
+    "enableAndGotoRemote": 3,
+    "enableAndLockoutLocal": 4,
+    "enableAndGTRLLO": 5,
+    "justGTL": 6,
 }
-REMOTELOCALCONTROLCODE.update(
-    {value: key for (key, value) in REMOTELOCALCONTROLCODE.items()}
-)
```
MatthieuDartiailh, line 106: Thanks for splitting the dicts

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
-    def send(self, data):
-        """Sends the data on the synchronous channel.
+    def send(self, data: bytes) -> int:
+        """Send the data on the synchronous channel.
 
         More than one packet may be necessary in order
-        to not exceed max_payload_size.
+        to not exceed _max_payload_size.
         """
+        # print(f"send({data=})")
```
MatthieuDartiailh, line 431: Look like debug leftover

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
 
         Terminate after max_len bytes or after receiving a DataEnd message
         """
 
-        # if we don't already have a Receiver object, create one
-        if self.receiver is None:
-            self.receiver = Receiver(self._sync, self.last_message_id)
-
-        # allocate a buffer and receive the data into it
+        # print(f"receive({max_len=})")
```
MatthieuDartiailh, line 456: debug cruft ?

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
+        # if we aren't already receiving, initialize the _expected_message_id
+        # and the payload length
+        if self._expected_message_id is None:
+            self._expected_message_id = self._last_message_id
+            # print(f"{self._expected_message_id=}")
```
MatthieuDartiailh, line 461:
```suggestion
```

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
+        while bytes_recvd < max_len:
+            if payload_remaining <= 0:
+                if msg_type == "DataEnd":
+                    # truncate to the actual number of bytes received
+                    recv_buffer = recv_buffer[:bytes_recvd]
+                    break
+                msg_type, payload_remaining = self._next_data_header()
+
+            request_size = min(payload_remaining, max_len - bytes_recvd)
+            # print(f"{payload_remaining=}, {request_size=}")
```
MatthieuDartiailh, line 484:
```suggestion
```

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
+            receive_exact_into(self._sync, view[:request_size])
+            payload_remaining -= request_size
+            bytes_recvd += request_size
+            view = view[request_size:]
+
+        if bytes_recvd > max_len:
+            raise MemoryError("scribbled past end of recv_buffer")
+
+        # print(f" -> {recv_buffer=}")
```
MatthieuDartiailh, line 493:
```suggestion
```

pyvisa_py/protocols/hislip.py
```diff
@@ -145,96 +393,165 @@ def __init__(self, ip_addr, open_timeout=None, port=PORT):
...
+                self._send_data_packet(data_view[: self._max_payload_size])
+                bytes_sent = self._max_payload_size
 
             data_view = data_view[bytes_sent:]
             num_bytes_to_send -= bytes_sent
 
         return len(data)
 
-    def receive(self, max_len=4096):
+    def receive(self, max_len: int = 4096) -> bytes:
```
MatthieuDartiailh, line 450: Does HiSLIP gives any warranty that when asked for 1 bytes it will return a single byte ? I know we have such an issue in VXI-11 where we get full messages and we need to buffer internally.

pyvisa_py/protocols/hislip.py
```diff
@@ -250,463 +567,179 @@ def initialize(self, version=(1, 0), vendor_id=b"xx", sub_address=b"hislip0"):
             vendor_id,
             len(sub_address),
         )
+        # txdecode(header, sub_address)
```
MatthieuDartiailh, line 570: ?

pyvisa_py/protocols/hislip.py
```diff
@@ -250,463 +567,179 @@ def initialize(self, version=(1, 0), vendor_id=b"xx", sub_address=b"hislip0"):
...
-        )
-        self._async.sendall(header)
-        response_header = receive_header(
-            self._async, expected_message_type="AsyncRemoteLocalResponse"
+        ctrl_code = REMOTELOCALCONTROLCODE[remotelocalcontrol]
+        send_msg(
+            self._async, "AsyncRemoteLocalControl", ctrl_code, self._last_message_id
         )
-        return response_header.server_status
+        _ = AsyncRemoteLocalResponse(self._async)
```
MatthieuDartiailh, line 645:
```suggestion
       AsyncRemoteLocalResponse(self._async)
```

pyvisa_py/protocols/hislip.py
```diff
@@ -250,463 +567,179 @@ def initialize(self, version=(1, 0), vendor_id=b"xx", sub_address=b"hislip0"):
...
+#     msg_type = MESSAGETYPE_STR[msg_type]
+#     print(
+#         f"Tx: {prologue=}, "
+#         f"{msg_type=}, "
+#         f"{control_code=}, "
+#         f"{message_parameter=}, "
+#         f"{payload_length=}, "
+#         f"{len(payload)=}, "
+#         f"{bytes(payload[:20]).decode('iso-8859-1')!r}"
+#     )
```
MatthieuDartiailh, line 745: Are those worth keeping around ? If so please add a comment explaining their purpose.

pyvisa_py/tcpip.py
```diff
@@ -64,10 +64,11 @@ def __new__(
         return newcls(resource_manager_session, resource_name, parsed, open_timeout)
 
 
-@Session.register(constants.InterfaceType.tcpip, "HISLIP")
 class TCPIPInstrHiSLIP(Session):
     """A TCPIP Session built on socket standard library using HiSLIP protocol."""
 
+    session_type = (constants.InterfaceType.tcpip, "HISLIP")
```
MatthieuDartiailh, line 70: From the VISA point of view the resource_class is INSTR I think

pyvisa_py/tcpip.py
```diff
@@ -234,10 +235,11 @@ def __init__(
             rpc.RawTCPClient.__init__(self, host, prog, vers, port, open_timeout)
 
 
-@Session.register(constants.InterfaceType.tcpip, "VXI11")
 class TCPIPInstrVxi11(Session):
     """A TCPIP Session built on socket standard library using VXI-11 protocol."""
 
+    session_type = (constants.InterfaceType.tcpip, "VXI11")
```
MatthieuDartiailh, line 241: Same here VISA only provide one attribute or so to discriminate between VXI11 and HiSLIP.

MatthieuDartiailh (changes requested):
The refactoring looks quite nice ! 

It seems to me some debug code found its way there and I left a couple of comments/questions but we are close to being good to go.

pyvisa_py/tcpip.py
bobmcnamara: Doesn't the session_type value need to be unique so that the attribute machinery will work?  The VISA attributes for HiSLIP aren't the same as those for VXI-11.  If we use INSTR for both VXI-11 and HiSLIP, wouldn't the HiSLIP attributes  appear on VXI-11 instances, where they wouldn't make any sense?

(the following is from https://documentation.help/VISA.NET/Using%20the%20TCPIP%20Interface.htm)

The HiSLIP protocol introduces three new VISA attributes for HiSLIP devices:

VI_ATTR_TCPIP_HISLIP_OVERLAP_EN  -  If VI_TRUE, enables overlap mode for HiSLIP. If VI_FALSE, the synchronous mode is used. Refer to [Synchronous and Overlap Modes](https://documentation.help/VISA.NET/Using%20the%20TCPIP%20Interface.htm#RTF320033003200300037003a00) for a description of these modes.
VI_ATTR_TCPIP_HISLIP_VERSION  -  Returns the ViVersion in use by the current HiSLIP session. For example, HiSLIP version 1.0 returns  a ViVersion value of 0x00100000.
VI_ATTR_TCPIP_HISLIP_MAX_MESSAGE_KB  -  Specifies the largest size HiSLIP message VISA will accept in units of kilobytes;  defaults to 1024 (a 1 MB maximum message size).

pyvisa_py/protocols/hislip.py
bobmcnamara: I've added a comment above those routines, but left them commented out in order to support Python versions <= 3.7.

They were really useful in figuring out why I couldn't talk to the Keysight scope after some minor changes.  I had left out the network order format specifier in a call to struct.pack().  So when I set the max message size to 16MB, it looked like I was setting it to 4GB.  Except the Keysight scope has a bug where it throws away the upper 32 bits of the 64-bit unsigned value.  So it decided I had set the max message size to 0, and it kept sending me packets with zero length payloads.

Once I realized that it was ignoring the upper 32 bits, I couldn't resist seeing if it was interpreting the remaining bits as signed or unsigned.  Sure enough, the scope application crashes if you give it a value with bit 31 set.

pyvisa_py/tcpip.py
MatthieuDartiailh: If you look at the implementation of attributes (https://github.com/pyvisa/pyvisa/blob/main/pyvisa/attributes.py) and the default behavior of sessions (https://github.com/pyvisa/pyvisa-py/blob/main/pyvisa_py/sessions.py#L620) you will see that:
- we only use valid resource class 
- their use simply allow to quickly discard unsupported attributes
- even if the test success we may still return StatusCode.error_nonsupported_attribute

As a consequence I still believe we should use INSTR for both VXI-11 and HiSLIP and simply provide different implementation for supported attributes.  

BTW do you plan to add support for any attribute as part of this PR ?

pyvisa_py/protocols/hislip.py
bobmcnamara: I've added this change into my latest commit.

pyvisa_py/tcpip.py
bobmcnamara: According to rule 5.1.16 (https://www.ivifoundation.org/docs/vpp43_2016-02-26.pdf), it's mandatory to support VI_ATTR_TCPIP_ADDR, VI_ATTR_TCPIP_HOSTNAME, VI_ATTR_TCPIP_IS_HISLIP, and VI_ATTR_TCPIP_DEVICE_NAME.

And according to rule 5.1.17, HiSLIP implementations must support VI_ATTR_TCPIP_PORT, VI_ATTR_TCPIP_NODELAY, VI_ATTR_TCPIP_KEEPALIVE, VI_ATTR_TCPIP_HISLIP_OVERLAP_EN, VI_ATTR_TCPIP_HISLIP_VERSION, and VI_ATTR_TCPIP_HISLIP_MAX_MESSAGE_KB.

I could use some pointers on how to implement these attributes.

I see another applicable rule, 5.1.31, which says that if no device name is specified, a VXI11 connection should be attempted, and if that fails, a HiSLIP connection should be attempted.  I'll have to think about how to implement that one.  Maybe that won't happen any time soon.

pyvisa_py/tcpip.py
bobmcnamara: I thought about this, tried your suggestion, and I agree that using "INSTR" for the resource class is the correct thing to do here.

bobmcnamara:
@MatthieuDartiailh I think I've made all the changes you requested.  Once I add support for the mandatory attributes (which I could use your help with, BTW), I think it will be ready to release.

pyvisa_py/tcpip.py
MatthieuDartiailh: I will look into the attributes. But I may not block your PR on it. PyVISA-py implement very few of them and I do not want to block your work on this. However a number of those are static values and should be easy.

MatthieuDartiailh:
Thanks @bobmcnamara 

I added support for some attributes (please check the hislip values, I tried to get them from your code but I may have misinterpreted things). I hardcoded the hislip version which does not really make sense but I have no idea how to query it from the instrument. I noticed that you establish the communication with version 1.0. I wrote 1.1 because of the title of your PR. Should we change it to 1.0 ?

I also tried to test your implementation against my fake Keysight instrument with which I can communicate over hislip using Keysight VISA and which report using the hislip protocol at version 1.0 (in hex 0x100000) and I got this error:
```
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa\pyvisa\highlevel.py", line 3284, in open_resource
    res.open(access_mode, open_timeout)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa\pyvisa\resources\resource.py", line 278, in open
    self.session, status = self._resource_manager.open_bare_resource(
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa\pyvisa\highlevel.py", line 3209, in open_bare_resource
    return self.visalib.open(self.session, resource_name, access_mode, open_timeout)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\highlevel.py", line 167, in open
    sess = cls(session, resource_name, parsed, open_timeout)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\tcpip.py", line 71, in __new__
    return newcls(resource_manager_session, resource_name, parsed, open_timeout)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\sessions.py", line 325, in __init__
    self.after_parsing()
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\tcpip.py", line 99, in after_parsing
    self.interface = hislip.Instrument(
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 384, in __init__
    init = self.initialize()
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 570, in initialize
    return InitializeResponse(self._sync)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 249, in __init__
    super().__init__(sock, "InitializeResponse")
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 207, in __init__
    self.header = receive_exact(sock, HEADER_SIZE)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 148, in receive_exact
    receive_exact_into(sock, recv_buffer)
  File "c:\users\matthieu.dartiailh\documents\coding\projects\pyvisa-py\pyvisa_py\protocols\hislip.py", line 164, in receive_exact_into
    data_len = sock.recv_into(view, request_size)
BlockingIOError: [WinError 10035] A non-blocking socket operation could not be completed immediately
```

bobmcnamara:
There’s a subtle error in my title…it should say “HiSLIP spec V1.1”.  For spec V1.1, the protocol version is specified as 1.0:

Protocol Version

The version of the protocol defined in this document is 1.0. In general the protocol version is not the same as the specification version. 

So yes, it should be changed to 1.0.

Sorry, I hadn’t tried the connection without specifying an open_timeout value.  In the case you tried, the open_timeout value being passed in is 0, which causes the failure.

I guess I should change this line:

open_timeout = open_timeout if open_timeout is not None else 1000

to this:

open_timeout = open_timeout or 1000

Or 2000?  Or 5000?

MatthieuDartiailh:
Thanks for clarifying. Did you check the other values I set ?

I will try playing with the open timeout and see to pipe a reasonable default.

Finally since the PR allowing to list VXI11 resources is in, I am wondering if we could provide an equivalent for hislip. Is it possible ?

bobmcnamara:
What I found is in this spec: https://www.lxistandard.org/members/Adopted%20Specifications/Latest%20Version%20of%20Standards_/LXI%20Version%201.6/LXI_HiSLIP_Extended_Function_1.3_2022-05-26.pdf

It says "Devices implementing the LXI HiSLIP Function shall advertise that they accept HiSLIP connections via the HiSLIP DNS-SD service announcement", and "HiSLIP DNS-SD service announcements shall use the mDNS service type name ‘_hislip._tcp’".

Using the command "dns-sd -B _hislip._tcp" in one of the labs identified dozens of instruments.  So it is possible.

There is a Python package (https://pypi.org/project/zeroconf/) that has already done all the heavy lifting to make this work.  Running the script below also identified the instruments:

#################################################
import time
from zeroconf import ServiceBrowser, ServiceListener, Zeroconf

class MyListener(ServiceListener):

    def update_service(self, zc: Zeroconf, type_: str, name: str) -> None:
        print(f"Service {name} updated")

    def remove_service(self, zc: Zeroconf, type_: str, name: str) -> None:
        print(f"Service {name} removed")

    def add_service(self, zc: Zeroconf, type_: str, name: str) -> None:
        info = zc.get_service_info(type_, name)
        print(f"Service {name} added, service info: {info}")

zeroconf = Zeroconf()
listener = MyListener()
browser = ServiceBrowser(zeroconf, "_hislip._tcp.local.", listener)

time.sleep(1)
zeroconf.close()
#################################################

Unfortunately that package has the LGPL license.  Will that be okay, since it only has to be imported?

MatthieuDartiailh:
The open_timeout is somewhat cryptic in the VISA spec. It actually only applies to the time given to the instrument to release a VISA lock not to the time after which to abort the connection attempt (section 4.3.3.2). I have had users confused by this previously. And it looks like the VXI-11 implementation does a poor job of respecting this too.

I think the usual value in other implementation is 5000 ms.

For the scanning dependency, I am always confused by Licensing, one easy way around it would be to make it an optional dependency like we did with psutil for scanning VXI-11 resources. However we should use a more explicit name for the option: hislip-discovery maybe

MatthieuDartiailh:
I also confirmed that I can open a hislip resource and run a simple query using your code.

bobmcnamara:
In pyvisa_py.highlevel.open(), the access_mode is ignored.  Since the open_timeout only applies to locks, it should be ignored as well:

If the accessMode parameter requests a lock,
then this parameter specifies the absolute time
period (in milliseconds) that the resource
waits to get unlocked before this operation
returns an error.

pyvisa_py/tcpip.py
bobmcnamara: I added get/set for keepalive and max_message_kb for HiSLIP.  While I was adding the keepalive, I noticed that the keepalive for VXI11 is Linux specific...it breaks on OSX (and presumably on Windows, but I haven't tried that).

See https://stackoverflow.com/questions/12248132/how-to-change-tcp-keepalive-timer-using-python-script

pyvisa_py/protocols/hislip.py
```diff
@@ -424,25 +422,50 @@ def timeout(self, val: float) -> None:
...
         self._async.settimeout(self._timeout)
 
+    @property
+    def max_msg_size(self) -> int:
+        """the maximum HiSLIP message size in bytes"""
+        return self._max_msg_size
+
+    @max_msg_size.setter
+    def max_msg_size(self, size: int) -> None:
+        """the maximum HiSLIP message size in bytes"""
```
MatthieuDartiailh, line 432:
```suggestion
```

pyvisa_py/protocols/hislip.py
```diff
@@ -424,25 +422,50 @@ def timeout(self, val: float) -> None:
...
+        self._max_msg_size = self.async_maximum_message_size(size)
+
+    @property
+    def keepalive(self) -> bool:
+        """returns the status of the TCP keepalive.  If a connection is dropped
+        as a result of “keepalives,” the error code VI_ERROR_CONN_LOST is
+        returned to current and subsequent I/O calls on the session"""
```
MatthieuDartiailh, line 439:
```suggestion
        """Status of the TCP keepalive.  
        
        Keepalive is on/off for both the sync and async sockets

        If a connection is dropped as a result of “keepalives,” the error code 
        VI_ERROR_CONN_LOST is returned to current and subsequent I/O 
        calls on the session.

        """
```

pyvisa_py/protocols/hislip.py
```diff
@@ -424,25 +422,50 @@ def timeout(self, val: float) -> None:
...
+        return self._keepalive
+
+    @keepalive.setter
+    def keepalive(self, keepalive: bool) -> None:
+        """turns keepalive on/off for both the sync and async sockets"""
```
MatthieuDartiailh, line 444:
```suggestion
```

Only the getter dosctring is visible for properties

pyvisa_py/tcpip.py
```diff
@@ -97,19 +98,86 @@ def after_parsing(self) -> None:
...
+        # self.attrs[ResourceAttribute.resource_manufacturer_id] = 4015
+        # self.attrs[ResourceAttribute.resource_manufacturer_name] = 'Rohde & Schwarz GmbH'
+        # self.attrs[ResourceAttribute.resource_spec_version] = 0x0050_0800
+        # self.attrs[ResourceAttribute.user_data] = 0
+        # self.attrs[ResourceAttribute.write_buffer_size] = 4096
+
+    def get_max_message_kb(
+        self, attribute: ResourceAttribute
+    ) -> Tuple[int, StatusCode]:
+        """returns the maximum HiSLIP message size in kilobytes"""
```
MatthieuDartiailh, line 162:
```suggestion
        """Get the maximum HiSLIP message size in kilobytes."""
```

pyvisa_py/tcpip.py
```diff
@@ -97,19 +98,86 @@ def after_parsing(self) -> None:
...
+        max_msg_size_kb = int(round(self.interface.max_msg_size / 1024))
+        return max_msg_size_kb, StatusCode.success
+
+    def set_max_message_kb(
+        self, attribute: ResourceAttribute, size_kb: int
+    ) -> StatusCode:
+        """sets the maximum HiSLIP message size in kilobytes"""
```
MatthieuDartiailh, line 169:
```suggestion
        """Set the maximum HiSLIP message size in kilobytes."""
```

pyvisa_py/tcpip.py
```diff
@@ -97,19 +98,86 @@ def after_parsing(self) -> None:
...
+            raise ValueError("size must be >= 1 kilobyte")
+
+        if size_kb > 0xFFFF_FFFF:
+            raise ValueError("size exceeds the range in the VISA spec")
+
+        self.interface.max_msg_size = int(round(size_kb * 1024))
+        return StatusCode.success
+
+    def get_keepalive(self, attribute: ResourceAttribute) -> Tuple[bool, StatusCode]:
+        """returns the status of the TCP keepalive"""
```
MatthieuDartiailh, line 180:
```suggestion
        """Is TCP keepalive enabled for the resource."""
```

pyvisa_py/tcpip.py
```diff
@@ -97,19 +98,86 @@ def after_parsing(self) -> None:
...
+        return self.interface.keepalive, StatusCode.success
+
+    def set_keepalive(
+        self, attribute: ResourceAttribute, keepalive: bool
+    ) -> StatusCode:
+        """turns TCP keepalive on/off for this connection"""
```
MatthieuDartiailh, line 186:
```suggestion
        """Turns TCP keepalive on/off for this connection."""
```

pyvisa_py/tcpip.py
```diff
@@ -1099,3 +1107,40 @@ def _set_attribute(
 
         """
         raise UnknownAttribute(attribute)
+
+
+def get_services(service_type, wait_time=0.1):
+    if zeroconf is None:
+        raise NotImplementedError(
+            "zeroconf package not found...try 'pip install zeroconf'"
+        )
```
MatthieuDartiailh, line 1116:
```suggestion
        raise NotImplementedError(
            "HiSLIP instrument discovery requires the zeroconf package to be installed... "
            "try 'pip install zeroconf'"
        )
```

MatthieuDartiailh: We could also add an optional dependencies in pyproject.toml.

`hislip-discovery= ["zeroconf"]`

MatthieuDartiailh (commented):
Looks great I just made some suggestions regarding mostly docstring formatting.

pyproject.toml
```diff
@@ -47,6 +47,7 @@ gpib-ctypes = ["gpib-ctypes>=0.3.0"]
 serial = ["pyserial>=3.0"]
 usb = ["pyusb"]
 psutil = ["psutil"]
+zeroconf = ["zeroconf"]
```
MatthieuDartiailh, line 50:
```suggestion
hislip-discovery = ["zeroconf"]
```

This will make it easier to discover I believe (could be worth adding to the docs). I should also change the psutil one.

MatthieuDartiailh:
Looks great. But it seems you need to run black because I messed up my suggestions. Sorry for that.

bobmcnamara:
It was a mypy complaint, not a black complaint.  Seems to be fixed now.

MatthieuDartiailh:
If you can rebase your VICP PR I will do my best to review it shortly

PR #332, opened 2022-08-09
Added VICP protocol

bobmcnamara:
Building on the changes that added the HiSLIP protocol, I added the VICP protocol making use of the pyvicp package I recently added to PyPI.  (That package can't be directly integrated into pyvisa because of its licensing terms.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,6 +71,8 @@ Optionally:
 - linux-gpib (to interface with gpib instruments, only on linux)
 - gpib-ctypes (to interface with GPIB instruments on Windows and Linux)
 - psutil (to discover TCPIP devices across multiple interfaces)
+- zeroconf (for mDNS based service discovery)
+- pyvicp (to enable the Teledyne LeCroy proprietary VICP protocol)
 
 
 Installation
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,7 @@ Optionally:
 - linux-gpib (to interface with gpib instruments, only on linux)
 - gpib-ctypes (to interface with GPIB instruments on Windows and Linux)
 - psutil (to discover TCPIP devices across multiple interfaces)
-- zeroconf (for mDNS based service discovery)
+- zeroconf (for HiSLIP and VICP devices discovery)
 - pyvicp (to enable the Teledyne LeCroy proprietary VICP protocol)
 
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -29,7 +29,7 @@ interface.
 
 Discovery of both HiSLIP  and VICP devices relies on `mDNS`_, which is a protocol for
 service discovery in a local area network.  To enable resource
-discovery for HiSLIP, you should install `zeroconf`_.
+discovery for HiSLIP and VICP, you should install `zeroconf`_.
 
 The TCP/IP VICP protocol (proprietary to Teledyne LeCroy) depends on
 the `pyvicp`_ package.  You should install this package if you need to
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -27,10 +27,13 @@ To discover VXI-11 devices on all network interfaces, please install
 `psutil`_. Otherwise, discovery will only occur on the default network
 interface.
 
-Discovery of HiSLIP devices relies on `mDNS`_, which is a protocol for
+Discovery of both HiSLIP  and VICP devices relies on `mDNS`_, which is a protocol for
 service discovery in a local area network.  To enable resource
 discovery for HiSLIP, you should install `zeroconf`_.
 
+The TCP/IP VICP protocol (proprietary to Teledyne LeCroy) depends on
+the `pyvicp`_ package.  You should install this package if you need to
+use VICP.
 
 
 Serial resources: ASRL INSTR
@@ -110,3 +113,4 @@ form GitHub_::
 .. _`psutil`: https://pypi.org/project/psutil/
 .. _`mDNS`: https://en.wikipedia.org/wiki/Multicast_DNS
 .. _`zeroconf`: https://pypi.org/project/zeroconf/
+.. _`pyvicp`: https://pypi.org/project/pyvicp/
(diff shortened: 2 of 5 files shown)


MatthieuDartiailh:
Why did you close the PR ?

bobmcnamara:
I dunno...maybe it happened when the cat walked across my laptop?

README.rst
```diff
@@ -71,6 +71,8 @@ Optionally:
 - linux-gpib (to interface with gpib instruments, only on linux)
 - gpib-ctypes (to interface with GPIB instruments on Windows and Linux)
 - psutil (to discover TCPIP devices across multiple interfaces)
+- zeroconf (for mDNS based service discovery)
```
MatthieuDartiailh, line 74:
```suggestion
- zeroconf (for HiSLIP and VICP devices discovery)
```

docs/source/installation.rst
```diff
@@ -27,10 +27,13 @@ To discover VXI-11 devices on all network interfaces, please install
 `psutil`_. Otherwise, discovery will only occur on the default network
 interface.
 
-Discovery of HiSLIP devices relies on `mDNS`_, which is a protocol for
+Discovery of both HiSLIP  and VICP devices relies on `mDNS`_, which is a protocol for
 service discovery in a local area network.  To enable resource
 discovery for HiSLIP, you should install `zeroconf`_.
```
MatthieuDartiailh, line 32:
```suggestion
discovery for HiSLIP and VICP, you should install `zeroconf`_.
```

pyproject.toml
```diff
@@ -48,6 +48,7 @@ serial = ["pyserial>=3.0"]
 usb = ["pyusb"]
 psutil = ["psutil"]
 hislip-discovery = ["zeroconf"]
+vicp = ["pyvicp"]
```
MatthieuDartiailh, line 51: Would it make sense to list zeroconf here also ?

pyvisa_py/tcpip.py
```diff
@@ -769,6 +778,215 @@ def _set_timeout(self, attribute: ResourceAttribute, value: int) -> StatusCode:
...
+    # for a specific kind of resource
+    parsed: rname.TCPIPInstr
+
+    @staticmethod
+    def list_resources(wait_time=1.0) -> List[str]:
+        resources = []
+        services = get_services("_lxi._tcp.local.", wait_time=wait_time)
+        for host, properties in services.items():
+            if properties["Manufacturer"].lower().startswith("lecroy"):
+                resources.append(f"TCPIP::{host}::VICP::INSTR")
```
MatthieuDartiailh, line 801: I may be misremembering but I had in mind that resource strings for VICP devices were of the form `VICP::{ip address}::INSTR`. I will double check.

MatthieuDartiailh (commented):
I left some minor comments but I need to check the usual resource string, your format does not match what I had in mind.

pyvisa_py/tcpip.py
bobmcnamara: The way they provided VICP support in NI-VISA struck me as a hack, making VICP support inconsistent with HiSLIP support.  You want to follow their lead?

MatthieuDartiailh: I simply worry that people will get confused if they try to follow their instrument manual. I often wonder how many people read pyvisa docs so I fear that if they need to reach to pyvisa-py doc to figure out why they cannot connect to their VICP device we will get a lot of reports of confused users. 

I agree with you on the principle but I fear it is not practical since Lecroy set the precedent a long time ago.

bobmcnamara: Yes, I see your point.  But I'm also up against a time constraint...I'm retiring in a couple of weeks and won't have access to instruments for testing after that.  I'll see what I can get done...

bobmcnamara: The changes using resource strings of the form `VICP::{ip address}::INSTR` are working using your vicp branch of pyvisa, but I don't know how to tell the continuous integration tools to use that branch.

bobmcnamara: Nevermind, found it: use "@" instead of "#".

MatthieuDartiailh:
Sorry I missed your latest comments. I saw the last commit but did not realize the rest was already done. Thanks again for your great contributions. I will try to merge both the PyVISA PR and this one quickly. I also found I could access an oscilloscope supporting VICP so I will be able to do investigate any issue that arises later on.

bobmcnamara:
From the LeCroy XStream manual:

`The COMM_HEADER command controls the way the oscilloscope formats responses to queries. There are three response formats: LONG, in which responses start with the long form of the header word; SHORT, where responses start with the short form of the header word; and OFF, for which headers are omitted from the response and units in numbers are suppressed.

Unless you request otherwise, the SHORT response format is used.`

I'm not sure why the response format is different depending on the protocol (VICP vs. VXI-11), but I always sent the command "CHDR OFF" to force it off.

bobmcnamara:
Here's a routine I've used to either reset the connection or to force VICP protocol when it wasn't convenient to physically access the machine:

```
def _steal_connection(ipaddr, port=80, protocol="vicp"):
    """
    LeCroy scopes only support a single VICP connection.  This procedure
    provides a way to free up that single connection.  In general, this
    would be used in instances where the connection didn't close cleanly
    (for instance, if you closed your laptop and went home).

    protocol is one of "vicp", "vxi11", or "usb".
    """
    from urllib.request import Request, urlopen
    from urllib.parse import urlencode

    url = "http://%s:%s/instrumentstatus.html" % (ipaddr, port)

    # turn remote interface off...
    data = urlencode(dict(Interface="Off"))
    req = Request(url, data.encode())
    response = urlopen(req)
    the_page = response.read()

    # turn remote interface back on...
    interface = dict(
        vicp="TCPIP\r\n(VICP)",
        vxi11="LXI\r\n(VXI11)",
        usb="USBTMC",
    )
    data = urlencode(dict(Interface=interface[protocol]))
    req = Request(url, data.encode())
    response = urlopen(req)
    the_page = response.read()

```

MatthieuDartiailh:
Thanks !

I got my hands on another scope. I will do my best to test again soon.

MatthieuDartiailh:
I never managed to open a VICP connection using the lecroy passport. I will merge https://github.com/pyvisa/pyvisa/pull/699 shortly and cut a release. I will then update this PR to bump the minimal pyvisa version, merge and release. Thanks again for your amazing contributions and sorry for the slow process.

PR #345, opened 2022-12-27
Adjusting license tag to use SPDX identifier

hyandell:
Hi Pyvisa folk,

My understanding is that, when possible, the license field ought to contain a descriptive identifier rather than link to the license. Typically that's when an SPDX license identifier exists. That makes it easier for automation, like the pypi website, to more cleanly describe the license.

As your license is the MIT license, here's a proposed PR to adjust it to be "MIT" rather than copying the whole license in.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "PyVISA-py"
 description = "Pure Python implementation of a VISA library."
 readme = "README.rst"
 requires-python = ">=3.7"
-license = {file = "LICENSE"}
+license = "MIT"
 authors = [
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
Your approach aligns with PEP 639 but it has not been accepted yet. Do you know if it is already supported by setuptools ?

hyandell:
Thanks for the heads up on PEP 639; I come to Python by way of licensing in the Java world so hadn't discovered that yet.

I haven't seen that show up in any PyPI packages, so I'm guessing it's not supported yet, or not known.

PR #354, opened 2023-02-15
fix(#346): don't advertize python 3.7 support

### LINKED ISSUE #346: python 3.8 syntax breaks 3.7 compatibility
https://github.com/pyvisa/pyvisa-py/blob/756152a97da3e7783d24b74fa09ae461eb92c787/pyvisa_py/tcpip.py#L1329

installing using pip3 install pyvisa-py on debian 10 (python 3.7.3) leads to no support for TCPIP

The above referenced line prevents the registration of TCP handlers TCPIPInstrSession, TCPIPSocketSession

sbor23:
Make sure package managers don't pull newer version for python 3.7

This should have happened in 0.6.0 but I don't think we can fix this. Probably pip will still try to install 0.6.0-0.6.2 for python 3.7,  but correctly these releases should be fixed with this patch.

- [x] Closes #346

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 name = "PyVISA-py"
 description = "Pure Python implementation of a VISA library."
 readme = "README.rst"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 license = {file = "LICENSE"}
 authors = [
   {name = "Hernan E. Grecco", email = "hernan.grecco@gmail.com"},


MatthieuDartiailh:
Thanks

sbor23:
Do you think this could be somehow backported to existing 0.6.x releases and push them to pypi?

sbor23:
> Do you think this could be somehow be backported to existing 0.6.x releases and push them to pypi?

@MatthieuDartiailh what do you think would that be possible? Because people will continue to get invalid versions 0.6.x otherwise.

MatthieuDartiailh:
The release 0.6.2 supports 3.7 so people should be fine. I could yank the earlier releases but people are unlikely to get them.

sbor23:
You're right, no need to yank older versions, but the problem is that 0.6.2 is broken too. So wee probably need first a version that advertizes python 3.7 and works, then we can make a new version that removes python 3.7.

Let's discuss in the issue instead: https://github.com/pyvisa/pyvisa-py/issues/355

PR #379, opened 2023-06-12
Support for GPIB secondary addressing

### LINKED ISSUE #307: secondary_address ignored
Current code ignores the user defined "secondary_address" and is hard coded to "0" in "gpib.py" line 258 "sad = 0"

Simple fix can be done by adding a line "sad = int(self.parsed.secondary_address) + 0x60" below line 264 "pad = int(self.parsed.primary_address)". 

"0x60" above is the secondary_address offset used by linux-gpib backend. Not sure what value it would be if Keysight or other backend is used:
https://linux-gpib.sourceforge.io/doc_html/reference-function-ibsad.html

drogenlied:
PyVISA 1.12 made proper support for GPIB secondary addressing possible by fixing address parsing.

This pull request enables secondary addressing and extends list_resources() to detect listeners on secondary addresses with minimal effort.

Tested  with a HP E1206A command module in a VXI chassis and several VXI modules, an Anritsu MS8604A spectrum analyzer and a Prema 6000 DMM on the same bus with an Agilent 82357B bus master driven by the linux-gpib Python module.

I see no opportunity to write unit tests as linux-gpib lacks a method to add mock devices and all changes interact directly with the linux-gpib API.

- [x] Closes #307 

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -35,6 +35,16 @@ from a pure primary address like `GPIB0::9::INSTR`.
 ``ResourceManager.list_resources()`` has become slower as a result,
 as it now needs to check 992 addresses per GPIB controller instead of just 31.
 
+For every primary address where no listener is detected, all
+secondary addresses are checked for listeners as well to find, for example,
+VXI modules controlled by an HP E1406A.
+
+For primary addresses where a listener is detected, no secondary addresses are
+checked as most devices simply ignore secondary addressing.
+
+If you have a device that reacts to the primary address and has different
+functionality on some secondary addresses, please leave a bug report.
+
 
 Can PyVISA-py be used from a VM?
 --------------------------------
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -21,6 +21,21 @@ The IVI compliant VISA implementations available (`National Instruments NI-VISA`
 certain systems. We wanted to provide a compatible alternative.
 
 
+Are GBIP secondary addresses supported?
+---------------------------------------
+
+GPIB secondary addresses are supported in NI-VISA fashion, meaning that the
+secondary address is not 96 to 126 as transmitted on the bus, but 0 to 30.
+
+For expample, `GPIB0::9::1::INSTR` is the address of the first VXI module
+controlled by a GPIB VXI command module set to primary address `9`, while
+the command module itself is found at `GPIB0::9::0::INSTR`, which is distinct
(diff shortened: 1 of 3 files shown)


MatthieuDartiailh:
This looks great ! I won't be able to test but the changes look reasonable and the heavy lifting is not done in pyvisa-py so I will trust you that things work.

Could add an entry to the changelog ? And update the docs to mention that secondary addresses are supported.

drogenlied:
Thanks for the encouragement, I hope that the added FAQ entry is acceptable for documentation.

As devices with secondary addresses are rare, I am not sure if the implemented behavior is as specified or whether we will ever find out.

docs/source/faq.rst
```diff
@@ -21,6 +21,21 @@ The IVI compliant VISA implementations available (`National Instruments NI-VISA`
...
+GPIB secondary addresses are supported in NI-VISA fashion, meaning that the
+secondary address is not 96 to 126 as transmitted on the bus, but 0 to 30.
+
+For expample, `GPIB0::9::1::INSTR` is the address of the first VXI module
+controlled by a GPIB VXI command module set to primary address `9`, while
+the command module itself is found at `GPIB0::9::0::INSTR`, which is distinct
+from a pure primary address like `GPIB0::9::INSTR`.
+
+``ResourceManager.list_resources()`` has become slower as a result,
+as it now needs to check 992 addresses per GPIB controller instead of just 31.
```
MatthieuDartiailh, line 36: Out of curiosity if there is no answer on a primary address is it possible to get a response with a secondary one ? If not it could help speed things up.

MatthieuDartiailh (approved):
Looks good. I have just one quick question regarding list resources.

docs/source/faq.rst
drogenlied: Yes that is possible, the E1406A GPIB command module does not answer on its primary address, only on the primary address with a correct secondary address set, which allows to enumerate all modules in the VXI chassis.

The other two devices I tested do not care about the secondary address, they answer on the primary and on the primary + any secondary address.

It should be possible for a device to answer on the primary **and** have subdevices on secondary addresses, but I have never heard of one, maybe the ICS 4896 GPIB to quad serial converter, its product brief is not very clear on the topic.

MatthieuDartiailh: Could you add a comment detailing the possible behavior so that nobody gets the idea to try to optimize things later ?

MatthieuDartiailh:
Thanks !

PR #383, opened 2023-06-23
fix(hislip): drop expected message id on read timeout

### LINKED ISSUE #376: TCP Socket connection: Issuing a query after a query that has timed out causes the previous result to be returned
If I issue a query that times out, and then issue a second query, that second query incorrectly returns the result from the first (timedout) query. The same test works fine when using the Keysight IVI backend.

To reproduce:
--------------
* Set instrument timeout to 5 seconds
* `*RST`
* `*OPC?`    <-- The *RST takes about 10 seconds so *OPC? times out after 5 seconds as expected
* Wait 20 seconds to ensure that the *RST has completed
* `IDN?`

The problem is that *IDN? returns '1' which is the result from the *OPC? query.
When using the Keysight IVI backend the *IDN? correctly returns 'Keysight Technologies,C8700200A Test Application Framework,US12345678,20.90.0.9010'

**Instrument details:**
* Keysight 5GNR Test Application Framework
* Communication: TCP Socket

**Code:**

```python
from datetime import datetime
import time
import pyvisa


def debug(*args):
    print(datetime.now(), *args)


rm = pyvisa.ResourceManager('@py')    # If I remove '@py' then it works i.e. *IDN? returns the instrument name
instrument = rm.open_resource('TCPIP0::127.0.0.1::5125::SOCKET')
instrument.timeout = 5000   # Small value to make *OPC? timeout
instrument.read_termination = '\n'

try:
    debug('Calling *RST')
    instrument.write('*RST')
    debug('Calling *OPC?')
    instrument.query('*OPC?')
except Exception as ex:
    debug(ex)

time.sleep(20)

debug('Calling *IDN?')
result = instrument.query('*IDN?')
debug('Result:', result)
```

**Incorrect result using pyvisa-py backend:**
```
2023-06-05 09:48:28.934000 Calling *RST 
2023-06-05 09:48:28.934000 Calling *OPC?
2023-06-05 09:48:33.989677 VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
2023-06-05 09:48:53.993221 Calling *IDN?
2023-06-05 09:48:53.995160 Result: 1   
```

**Correct result using Keysight IVI backend (simply removing '@py' from the ResourceManager constructor):**
```
2023-06-05 09:55:28.857054 Calling *RST 
2023-06-05 09:55:28.858046 Calling *OPC?
2023-06-05 09:55:33.881024 VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
2023-06-05 09:55:53.892010 Calling *IDN?
2023-06-05 09:55:53.894488 Result: Keysight Technologies,C8700200A Test Application Framework,US12345678,20.90.0.9010
```

**Output of pyvisa-info**
```
Machine Details:
   Platform ID:    Windows-10-10.0.19044-SP0
   Processor:      Intel64 Family 6 Model 141 Stepping 1, GenuineIntel

Python:
   Implementation: CPython
   Executable:     C:\Program Files\Python38\python.exe
   Version:        3.8.16
   Compiler:       MSC v.1900 64 bit (AMD64)
   Bits:           64bit
   Build:          Dec 20 2022 23:15:53 (#default)
   Unicode:        UCS4

PyVISA Version: 1.13.0

Backends:
   ivi:
      Version: 1.13.0 (bundled with PyVISA)
      #1: C:\Windows\system32\visa32.dll:
         found by: auto
         bitness: 64
         Vendor: Keysight Technologies
         Impl. Version: 1377199637
         Spec. Version: 5244928
      #2: C:\Windows\system32\visa64.dll:
         found by: auto
         bitness: 64
         Vendor: Keysight Technologies
         Impl. Version: 1377199637
         Spec. Version: 5244928
   py:
      Version: 0.7.0
      TCPIP INSTR: Available
         Resource discovery:
         - VXI-11: partial (psutil not installed)
      TCPIP SOCKET: Available
      ASRL INSTR:
         Please install PySerial (>=3.0) to use this resource type.
         No module named 'serial'
      USB INSTR:
         No module named 'usb'
      USB RAW:
         No module named 'usb'
      VICP INSTR:
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
      GPIB INTFC:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
```

arr-ee:
As per #376, current implementation was not resetting `expected_message_id` in case of timeout, potentially leading to stale responses being returned to the caller.

This is very much a bandaid while I'm working on both better integration tests and on a (hopefully) more robust hislip implementation.

- [x] Closes #376 (insert issue number if relevant)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ PyVISA-py Changelog
 - addd URL-support to ASLR devices PR #386
 - add support for GPIB secondary addresses
 - fix missing sock.close() in rpc _connect()
+- fix HiSLIP message tracking after read timeout #376
 
 0.7.0 (05/05/2023)
 ------------------
--- a/pyvisa_py/protocols/hislip.py
+++ b/pyvisa_py/protocols/hislip.py
@@ -400,7 +400,6 @@ class Instrument:
         self.keepalive = False
         self.timeout = timeout
         self._rmt = 0
-        self._expected_message_id: Optional[int] = None
         self._message_id = 0xFFFF_FF00
         self._last_message_id: Optional[int] = None
         self._msg_type: str = ""
@@ -434,6 +433,18 @@ class Instrument:
     @max_msg_size.setter
     def max_msg_size(self, size: int) -> None:
         self._max_msg_size = self.async_maximum_message_size(size)
+
+    @property
+    def last_message_id(self) -> Optional[int]:
+        return self._last_message_id
+
+    @last_message_id.setter
+    def last_message_id(self, message_id: Optional[int]) -> None:
+        """Re-set last message id and related attributes"""
+        self._last_message_id = message_id
+        self._rmt = 0
+        self._payload_remaining = 0
+        self._msg_type = ""
 
     @property
     def keepalive(self) -> bool:
@@ -487,12 +498,6 @@ class Instrument:
         """
 
         # print(f"receive({max_len=})")  # uncomment for debugging
-        # if we aren't already receiving, initialize the _expected_message_id
-        # and the payload length
-        if self._expected_message_id is None:
-            self._expected_message_id = self._last_message_id
-            self._msg_type = ""
-            self._payload_remaining = 0
 
         # receive data, terminating after len(recv_buffer) bytes or
         # after receiving a DataEnd message.
@@ -521,8 +526,7 @@ class Instrument:
         if bytes_recvd > max_len:
             raise MemoryError("scribbled past end of recv_buffer")
 
-        # if there is no data remaining, set the RMT flag and set the
-        # _expected_message_id to None
+        # if there is no data remaining, set the RMT flag
         if self._payload_remaining == 0 and self._msg_type == "DataEnd":
             #
             # From IEEE Std 488.2: Response Message Terminator.
@@ -532,7 +536,6 @@ class Instrument:
             # this is implied by the DataEND message.
             #
             self._rmt = 1
-            self._expected_message_id = None
 
         return recv_buffer
 
@@ -555,7 +558,7 @@ class Instrument:
 
                 if (
                     header.message_parameter == 0xFFFF_FFFF
-                    or header.message_parameter == self._expected_message_id
+                    or header.message_parameter == self.last_message_id
                 ):
                     break
 
@@ -655,7 +658,7 @@ class Instrument:
         #     C->S: AsyncLock
         #     S->C: AsyncLockResponse
         ctrl_code = LOCKCONTROLCODE["release"]
-        send_msg(self._async, "AsyncLock", ctrl_code, self._last_message_id)
+        send_msg(self._async, "AsyncLock", ctrl_code, self.last_message_id)
         response = AsyncLockResponse(self._async)
         return response.lock_response
 
@@ -668,7 +671,7 @@ class Instrument:
         #     S->C: AsyncRemoteLocalResponse
         ctrl_code = REMOTELOCALCONTROLCODE[remotelocalcontrol]
         send_msg(
-            self._async, "AsyncRemoteLocalControl", ctrl_code, self._last_message_id
+            self._async, "AsyncRemoteLocalControl", ctrl_code, self.last_message_id
         )
         AsyncRemoteLocalResponse(self._async)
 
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
@bobmcnamara could you comment since you contributed the initial hislip support ?

bobmcnamara:
This looks good to me, though I think it would be better to use "raise" rather than "raise e".

MatthieuDartiailh:
Thanks @bobmcnamara 

I agree with you a bare raise would make more sense. @arr-ee I will merge once you make the change.

bobmcnamara:
@arr-ee Could you also add a comment in the except branch?  Maybe something like "force the next call to re-initialize"?

arr-ee:
> it would be better to use "raise" rather than "raise e".

TIL! Agreed.

> Maybe something like "force the next call to re-initialize"?

So I went to do this, and decided to re-read the spec since I remembered something about discarding buffered messages there. I was also bothered by the fact that `read`ing after timeouted command actually responded would discard that response now even if we have not sent any commands yet — I do not think this is desirable.

Turns out, the [spec](https://ivifoundation.org/downloads/Protocol%20Specifications/IVI-6.1_HiSLIP-2.0-2020-04-23.pdf) agrees with current me and not with my original approach (from `3.1.2 Synchronized Mode Client Requirements`):

> 3. When the client sends Data, DataEND or Trigger, if there are any whole or partial server messages that have been validated per rules 1 and 2 and buffered, they shall be cleared.

In our case we are not doing any buffering in the sense they're using the word (== we're not eagerly reading out the socket buffer and process messages, storing them for consumption later), so what I think this means for us is that `_expected_message_id` must be set when sending these packets, and only unset once we have successfully processed a response that matches `_expected_message_id`.

What I'm not sure about is this bit: https://github.com/pyvisa/pyvisa-py/blob/c33f4f200dbbcbcd5132d5de61c570acf9c77581/pyvisa_py/protocols/hislip.py#L490-L495

As far as I understand, this 

1. is only needed to reset stateful fields related to the transfer
2. is always executed _unless_ response is larger than `max_len`
3. leaves us in a bad state if we have not read out the entirety of the payload before making another message+read (we'd go straight to trying to read the "remainder" of the initial payload)

My proposal is to remove this entire block, and effectively move its body into a separate method called when we send `Data`/`DataEnd`/`Trigger` messages as per spec. I smoke-tested this, and it passes both the simulated test from the issue, as well as a test against a real instrument. It does have a higher chance of introducing breakages (I think) compared to the solution in this PR, but imo the issue it fixes is more problematic.

MatthieuDartiailh:
Seems reasonable to me but I am not familiar with hislip. @bobmcnamara your input would be valuable here.

bobmcnamara:
@MatthieuDartiailh This is one of those cases when after looking at the code and then the spec, I'm left wondering what I was thinking when I wrote that code.

@arr-ee I think that the _expected_message_id can be deleted, and the comparison in the _next_data_header method can be to _last_message_id.  Then the entire block https://github.com/pyvisa/pyvisa-py/blob/c33f4f200dbbcbcd5132d5de61c570acf9c77581/pyvisa_py/protocols/hislip.py#L490-L495 can be eliminated.  And I think you'll have to move the self._msg_type = "" down where self._rmt = 1.

I no longer have access to machines to test any of this, so I might have missed something.

arr-ee:
> I think that the _expected_message_id can be deleted, and the comparison in the _next_data_header method can be to _last_message_id

Fair, it appears we can just do that.

> Then the entire block can be eliminated.

My main concern there is the `_payload_remaining` that we might not reset correctly in the case of e.g. not receiving the entirety of the payload for some reason. I'd rather keep it as is for now until I write some tests.

bobmcnamara:
> My main concern there is the _payload_remaining that we might not reset correctly in the case of e.g. not receiving the entirety of the payload for some reason. I'd rather keep it as is for now until I write some tests.

Initially I had the same concern.  But _payload_remaining gets set to 0 in __init__() when the instance is created.  So for a quick fix, I don't think anything else is required.

For a more robust fix, I think we'd want to test _payload_remaining in trigger(), _send_data_packet(), and _send_data_end_packet().   If it is nonzero, call receive_flush() and then reset _payload_remaining.

I think that's the behavior called out in the spec.

MatthieuDartiailh:
Thanks @bobmcnamara for you input.

@arr-ee do you have the time to make the changes made suggested by @bobmcnamara ?

arr-ee:
Apologies for the delay, had to prioritise other stuff.

I have updated the PR, it's now https://github.com/pyvisa/pyvisa-py/pull/383#issuecomment-1614926835 + `_expected_message_id can be deleted, and the comparison in the _next_data_header method can be to _last_message_id` suggestion.

I am not comfortable dropping payload tracking reset just now, but I plan to revisit hislip support in the future.

MatthieuDartiailh:
Sounds good to me. I will merge in a week in case anybody else wants to comment.

PR #386, opened 2023-07-20
replace the call to serial.Serial with serial.serial_for_url

### LINKED ISSUE #385: [COM] Communication issue loop
For tests (in pymeasure) I would like to create a Visa-resource pointing to a loop as supported by PySerial.
Is there away to create a VISA resource string point to a `serial.serial_for_url("loop://", timeout=0.2)` object?

Or would there be another way to create a rm-object for a loop device?

------------------
* Link to the documentation (if available): https://pyserial.readthedocs.io/en/latest/url_handlers.html#loop

Output of `pyvisa-info`
-----------------------

```
Machine Details:
   Platform ID:    Linux-6.4.3-arch1-2-x86_64-with-glibc2.37
   Processor:      

Python:
   Implementation: CPython
   Executable:     /usr/bin/python
   Version:        3.11.3
   Compiler:       GCC 13.1.1 20230429
   Architecture:   ('x86', 64)
   Build:          Jun  5 2023 09:32:32 (#main)
   Unicode:        UCS4

PyVISA Version: 1.13.1.dev29+g55f01db

Backends:
   ivi:
      Version: 1.13.1.dev29+g55f01db (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.7.0
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR: Available via PyUSB (1.2.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.2.1). Backend: libusb1
      TCPIP INSTR: Available 
         Resource discovery:
         - VXI-11: ok
         - hislip: ok
      TCPIP SOCKET: Available 
      GPIB INSTR: Available via Linux GPIB (b'4.3.5')
      GPIB INTFC: Available via Linux GPIB (b'4.3.5')
      VICP INSTR:
```

LongnoseRob:
to support loop:// and other url-types supported in pyserial (>3.0)

tested with the follwing code on python 3.11.3, pyserial 3.5:
```
import pyvisa
abc = rm.open_resource("ASRLloop://::INSTR")
abc.timeout = 3000
abc.read_termination = "\r"
abc.write_termination = "\r\n"
abc.write("Test1234567890")
print(abc.read())
```

- [ ] Closes #385

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,9 +32,12 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
+      echo Install support packages
+      pip install pyserial
       echo Install project
       pip install git+https://github.com/pyvisa/pyvisa.git#egg=pyvisa
       pip install -e .
+      
     displayName: "Install dependencies"
 
   - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,11 +32,9 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
-      echo Install support packages
-      pip install pyserial
-      echo Install project
+      echo Install project amd required dependencies
       pip install git+https://github.com/pyvisa/pyvisa.git#egg=pyvisa
-      pip install -e .
+      pip install -e . poyserial
       
     displayName: "Install dependencies"
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,9 +32,9 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
-      echo Install project amd required dependencies
+      echo Install project and required dependencies
       pip install git+https://github.com/pyvisa/pyvisa.git#egg=pyvisa
-      pip install -e . poyserial
+      pip install -e .[serial]
       
     displayName: "Install dependencies"
 
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.7.1 (unreleased)
 ------------------
 
+- addd URL-support to ASLR devices PR #386
 - add support for GPIB secondary addresses
 
 0.7.0 (05/05/2023)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -35,6 +35,16 @@ Currently Pyvisa-py support the following resources:
 - USB INSTR
 - USB RAW
 
+    Note:
+    ASRL INSTR supports also URL Handlers like 
+    
+    - loop:// --> ASLRloop://::INSTR
+    - socket:// --> ASRLsocket://::INSTR
+
+    These entries will not be listed during the device discovery `rm.list_resources()`.
+    For further details see https://pyserial.readthedocs.io/en/latest/url_handlers.html
+    
+
 You can report a problem or ask for features in the `issue tracker`_.
 Or get the code in GitHub_.
 
--- a/pyvisa_py/serial.py
+++ b/pyvisa_py/serial.py
@@ -84,10 +84,8 @@ class SerialSession(Session):
         return "via PySerial (%s)" % ver
 
     def after_parsing(self) -> None:
-        cls = serial.Serial
-
-        self.interface = cls(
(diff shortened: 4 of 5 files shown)


MatthieuDartiailh:
Could add an entry to the docs ? Otherwise this looks good since we require pyserial >= 3.0 and this feature was introduced in 2.5

LongnoseRob:
> Could add an entry to the docs ? Otherwise this looks good since we require pyserial >= 3.0 and this feature was introduced in 2.5

Added some documentation, hope this is what you were thinking about.

What I am not sure about is how to add a test

LongnoseRob:
> Thanks for the doc update I believe it will do. My only remaining concern is the behavior one gets if pyvisa attempts to read/write an attribute which is not supported for a URL. I did not dive deep enough in pyserial docs to have a clear sense of what would happen then.

If it try an `assert_trigger() ` on a loop:// resource, this causes a `NotImplementedError`
If I try a `read_stb()`, this causes a `VisaIOError: VI_ERROR_NSUP_OPER (-1073807257): The given session or object reference does not support this operation.`

> As for testing since you can talk to the loop back you could mock a resource no ?

WIll work on that next

LongnoseRob:
added a unit test, but it seems i cannot trigger the test(s) from the cli:
```
~/builds/pyvisa-py$ python -m unittest pyvisa_py/testsuite/test_serial.py                                                                           |

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK

```

MatthieuDartiailh:
I may have switched to pytest at one point I will have to double check.

LongnoseRob:
> I may have switched to pytest at one point I will have to double check.

pytest seems to work (and also pointed out a deprecatuion issue..

MatthieuDartiailh:
CIs are not good. You may need to skip the new tests in some cases (note I did not check the exact failure).

LongnoseRob:
> CIs are not good. You may need to skip the new tests in some cases (note I did not check the exact failure).

The CI failed due to me using `.append` instead of directly setting the value.

MatthieuDartiailh:
I am fine with you adding to the CI env but I would also like to see the test behind a conditional skip if pyserial is not available.

LongnoseRob:
> I am fine with you adding to the CI env but I would also like to see the test behind a conditional skip if pyserial is not available.

- [x] conditional skip included
- [x] enviroment setup for tests includes now pyserial

azure-pipelines.yml
```diff
@@ -32,9 +32,12 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
+      echo Install support packages
+      pip install pyserial
       echo Install project
       pip install git+https://github.com/pyvisa/pyvisa.git#egg=pyvisa
       pip install -e .
```
MatthieuDartiailh, line 39: You could do `pip install -e .[serial]` instead with a comment explining why we install this extra.

MatthieuDartiailh (commented):
Looks good. I just have a nitpick regarding the CI configuration.

azure-pipelines.yml
LongnoseRob: If you would prefer it that way, I will change it.

azure-pipelines.yml
```diff
@@ -32,11 +32,9 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
-      echo Install support packages
-      pip install pyserial
-      echo Install project
+      echo Install project amd required dependencies
       pip install git+https://github.com/pyvisa/pyvisa.git#egg=pyvisa
-      pip install -e .
+      pip install -e . poyserial
```
MatthieuDartiailh, line 37:
```suggestion
      pip install -e .[serial]
```

azure-pipelines.yml
```diff
@@ -32,11 +32,9 @@ steps:
       source $HOME/miniconda3/bin/activate
       echo Activate environment
       call conda activate test_
-      echo Install support packages
-      pip install pyserial
-      echo Install project
+      echo Install project amd required dependencies
```
MatthieuDartiailh, line 35:
```suggestion
      echo Install project and required dependencies
```

MatthieuDartiailh (changes requested):

LongnoseRob:
Thanks for the review.

PR #388, opened 2023-07-27
Add missing sock.close() to rpc _connect()

ACharlyR:
The socket was not closed if the _connect() function timed out. This lead to an unclosed socket every time a second instance of pyvisa attempted to connect to a device while the first instance of pyvisa still had an open connection.

With this fix, the second instance can back off and try again at a later time without reaching an irrecoverable state.


MatthieuDartiailh (approved):
Thanks !

PR #405, opened 2023-12-13
Update index.rst

jackparmer:
Add a link to Flojoy project (PyVISA-py GUI)

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -58,3 +58,4 @@ Or get the code in GitHub_.
 .. _GitHub: https://github.com/pyvisa/pyvisa-py
 .. _`issue tracker`: https://github.com/pyvisa/pyvisa-py/issues
 
+If looking for a no-code GUI for PyVISA-py test automation, please see the [Flojoy project](https://github.com/flojoy-ai/studio)


arr-ee:
Would you care to elaborate?

jackparmer:
Sure thing. Flojoy is building a drag & drop GUI for interfacing with instruments through PyVISA-Py (or PySerial). Examples of instrument connectors already built using PyVISA-Py can be found here: https://docs.flojoy.ai/blocks/hardware/. It seems like there could be something like a "Projects using PyVISA-Py" section on the homepage to help visitors navigate what's available in the growing Python T&M ecosystem.

arr-ee:
There is precious few projects using pyvisa-py, since it doesn’t make sense to use directly — surely you know that.

Seeing a zero-effort advertising PR for a soon-to-turn-non-free project promising “gui” over same old unsafe and non-reusable patterns of access to instruments does exactly fill me with excitement — but that’s just me, of course.

MatthieuDartiailh:
I am not opposed to providing some exposure to flowjoy (since it may benefit PyVISA), however I agree that making it at the level of pyvisa-py makes no sense since it is merely a backend for pyvisa and that I doubt flowjot is re-inventing the PyVISA layer.

Issue #414, opened 2024-02-09
[COM] Communication issue with RIGOL oscilloscope (model DS1104z Plus) using pyvisa-py


Fellypao:
Instrument details
------------------
* Model: RIGOL DS1104z Plus
* Communication: USB
* Link to the documentation: https://www.batronix.com/files/Rigol/Oszilloskope/_DS&MSO1000Z/MSO_DS1000Z_ProgrammingGuide_EN.pdf

Output of `pyvisa-info`
-----------------------
Machine Details:
   Platform ID:    Linux-5.15.38-x86_64-AMD_Ryzen_5_3500U_with_Radeon_Vega_Mobile_Gfx-with-glibc2.33
   Processor:      AMD Ryzen 5 3500U with Radeon Vega Mobile Gfx

Python:
   Implementation: CPython
   Executable:     /home/fellype/progs_wo_installer/spyder/bin/python
   Version:        3.9.18
   Compiler:       GCC 11.2.0
   Bits:           64bit
   Build:          Sep 15 2023 13:03:38 (#main)
   Unicode:        UCS4

PyVISA Version: 1.13.0

Backends:
   ivi:
      Version: 1.13.0 (bundled with PyVISA)
      #1: /usr/lib64/libvisa.so.0.0.0:
         found by: auto
         bitness: 64
         Could not get more info:
            VI_ERROR_NSUP_ATTR (-1073807331): The specified attribute is not defined or supported by the referenced object.
   
   py:
      Version: 0.7.1
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR: Available via PyUSB (1.2.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.2.1). Backend: libusb1
      TCPIP INSTR: Available 
         Resource discovery:
         - VXI-11: ok
         - hislip: disabled (zeroconf not installed)
      TCPIP SOCKET: Available 
      VICP INSTR:
         Please install PyVICP to use this resource type.
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
      GPIB INTFC:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'

Problem description:
------------------
[...]

PR #417, opened 2024-03-06
Fix to use endpoints MaxPacketSize instead of hardcoded RECV_CHUNK size

sessl3r:
The issue was found to be not able to download screenshots from a RigolDS1074Z.
From this [issue](https://github.com/pyvisa/pyvisa/issues/481#issuecomment-1477334673) it was seen there is some workaround but not a fix.

As the same issue existed in the [linux-usbtmc](https://github.com/dpenkler/linux-usbtmc/issues/8) driver module it was easy to refix it here.

The main point is that those scopes have a limited MaxPacketSize which is reported by the USB device correctly and the TMC protocol is relying on using this MaxPacketSize.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,11 @@
 PyVISA-py Changelog
 ===================
 
+0.7.2 (unreleased)
+------------------
+
+- fix usbtmc to use MaxPacketSize reported by endpoint
+
 0.7.1 (26/10/2023)
 ------------------
 
--- a/pyvisa_py/protocols/usbtmc.py
+++ b/pyvisa_py/protocols/usbtmc.py
@@ -298,9 +298,6 @@ class USBRaw(object):
 
 
 class USBTMC(USBRaw):
-    # Maximum number of bytes per transfer (for sending and receiving).
-    RECV_CHUNK = 1024**2
-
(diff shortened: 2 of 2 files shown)


MatthieuDartiailh:
Wow a fix that helps with Rigol devices is definitively welcome ! The formatting failure is unrelated to your changes so do not worry about it. Could you add a changelog entry ? Once it is done I will merge.

@Fellypao it may be worth checking if this fixes your issue.

MatthieuDartiailh:
Perfect ! Thanks a ton !

Fellypao:
> @Fellypao it may be worth checking if this fixes your issue.
> 

@MatthieuDartiailh Yes, this definitely fixes my issue (#414). Thank you. And many thanks to @sessl3r :+1:

MatthieuDartiailh:
Amazing !!! I will try to cut a release ASAP

MatthieuDartiailh:
Could you open an issue so that the information is less likely to be lost ? 

Honestly I take the win of people having Rigol scope working out of the box over maximal performance for Agilent scope.

MatthieuDartiailh:
Thanks @sessl3r

PR #423, opened 2024-04-07
expand info/debug/warning msgs if USB device `serial_number` not readable

jllanfranchi:
if INSTR USB device,
- say that device is USB INSTR that failed in warning message
- catch exception
- debug message with traceback if logging level is DEBUG
- info message 1-line error summary if logging level is INFO
- if path f"/dev/bus/usb/{dev.bus:03d}/{dev.address:03d}" exists, check permissions on that file & refer to FAQ link (known issue on Linux)

if RAW USB device,
- say that device is USB RAW that failed in warning message

I will work on documentation fixes now, but wanted to get this in front of you for feedback in the meantime, @arr-ee (et al.).

- [ ] Working to close pyvisa issue [#758](https://github.com/pyvisa/pyvisa/issues/758)
- [x] Executed ``black . && isort -c . && flake8`` with ~no~ errors black & isort OK; flake8 doesn't like a few lines' lengths
- [ ] The change is fully covered by automated unit tests (TODO; feedback for what this could be?)
- [ ] Documented in docs/ as appropriate (TODO)
- [ ] Added an entry to the CHANGES file (TODO)

### DIFF
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -8,6 +8,9 @@
 """
 
 import errno
+import logging
+import os
+import traceback
 from typing import Any, List, Tuple, Type, Union
 
 from pyvisa import attributes, constants
@@ -277,9 +280,9 @@ class USBInstrSession(USBSession):
 
             try:
                 serial = dev.serial_number
-            except (NotImplementedError, ValueError):
+            except (NotImplementedError, ValueError) as err:
                 msg = (
-                    "Found a device whose serial number cannot be read."
+                    "Found a USB INSTR device whose serial number cannot be read."
                     " The partial VISA resource name is: " + fmt
                 )
                 logger.warning(
@@ -292,6 +295,39 @@ class USBInstrSession(USBSession):
                         "usb_interface_number": intfc,
                     },
                 )
+                logging_level = logger.getEffectiveLevel()
+                if logging_level <= logging.DEBUG:
+                    if exc_strs := traceback.format_exception(err):
+                        msg = "Traceback:"
+                        logger.debug(msg)
+                        logger.debug("-" * len(msg))
+                        for exc_str in exc_strs:
+                            for line in exc_str.split("\n"):
+                                logger.debug(line)
+                        logger.debug("-" * len(msg))
+                elif logging_level <= logging.INFO:
+                    if exc_strs := traceback.format_exception_only(err):
+                        logger.info(
+                            "Error raised from underlying module (pyusb): %s",
+                            exc_strs[0].strip(),
+                        )
+
+                dev_path = f"/dev/bus/usb/{dev.bus:03d}/{dev.address:03d}"
+                if os.path.exists(dev_path) and not os.access(dev_path, os.O_RDWR):
+                    missing_perms = []
+                    if not os.access(dev_path, os.O_RDONLY):
+                        missing_perms.append("read from")
+                    if not os.access(dev_path, os.O_WRONLY):
+                        missing_perms.append("write to")
+                    missing_perms_str = " or ".join(missing_perms)
+                    logger.warning(
+                        "User does not have permission to %s %s, so the above USBTMC"
+                        " device cannot be used by pyvisa; see"
+                        " https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/faq.html"
+                        " for more info.",
+                        missing_perms_str,
+                        dev_path,
+                    )
                 continue
 
             out.append(
@@ -336,7 +372,7 @@ class USBRawSession(USBSession):
                 serial = dev.serial_number
             except (NotImplementedError, ValueError, usb.USBError):
                 msg = (
-                    "Found a device whose serial number cannot be read."
+                    "Found a USB RAW device whose serial number cannot be read."
                     " The partial VISA resource name is: " + fmt
                 )
                 logger.warning(


jllanfranchi:
Examples of messages running the code

```python
import logging
import pyvisa

logging.basicConfig(level=logging.DEBUG)
rm.list_resources()
```

on Linux with permissions /dev/usb/001/016 not set correctly (I also have a USB RAW device on my system, so you can see the modified message for that as well).

* logging level set to WARN:

```
WARNING:pyvisa:Found a USB INSTR device whose serial number cannot be read. The partial VISA resource name is: USB0::4883::32888::???::0::INSTR
WARNING:pyvisa:User does not have permission to write to /dev/bus/usb/001/016, so the above USB INSTR device cannot be used by pyvisa; see https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/faq.html for more info.
WARNING:pyvisa:Found a USB RAW device whose serial number cannot be read. The partial VISA resource name is: USB0::1027::24593::???::0::RAW
```

* logging level set to INFO:

```
WARNING:pyvisa:Found a USB INSTR device whose serial number cannot be read. The partial VISA resource name is: USB0::4883::32888::???::0::INSTR
INFO:pyvisa:Error raised from underlying module (pyusb): ValueError: The device has no langid (permission issue, no string descriptors supported or device error)
WARNING:pyvisa:User does not have permission to write to /dev/bus/usb/001/016, so the above USB INSTR device cannot be used by pyvisa; see https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/faq.html for more info.
WARNING:pyvisa:Found a USB RAW device whose serial number cannot be read. The partial VISA resource name is: USB0::1027::24593::???::0::RAW
```

* with logging level set to DEBUG:

```
DEBUG:pyvisa:SerialSession was correctly imported.
DEBUG:pyvisa:USBSession and USBRawSession were correctly imported.
DEBUG:pyvisa:TCPIPSession was correctly imported.
DEBUG:pyvisa:GPIBSession was not imported No module named 'gpib'.
DEBUG:pyvisa:Created library wrapper for py
DEBUG:pyvisa:Created ResourceManager with session 1292698
WARNING:pyvisa:Found a USB INSTR device whose serial number cannot be read. The partial VISA resource name is: USB0::4883::32888::???::0::INSTR
DEBUG:pyvisa:Traceback:
DEBUG:pyvisa:----------
DEBUG:pyvisa:Traceback (most recent call last):
DEBUG:pyvisa:
DEBUG:pyvisa:  File "/home/justinl/src/pyvisa-py/pyvisa_py/usb.py", line 282, in list_resources
DEBUG:pyvisa:    serial = dev.serial_number
DEBUG:pyvisa:
DEBUG:pyvisa:  File "/home/justinl/miniconda3/envs/py310/lib/python3.10/site-packages/usb/core.py", line 864, in serial_number
DEBUG:pyvisa:    self._serial_number = util.get_string(self, self.iSerialNumber)
DEBUG:pyvisa:
DEBUG:pyvisa:  File "/home/justinl/miniconda3/envs/py310/lib/python3.10/site-packages/usb/util.py", line 313, in get_string
DEBUG:pyvisa:    raise ValueError("The device has no langid"
DEBUG:pyvisa:
DEBUG:pyvisa:ValueError: The device has no langid (permission issue, no string descriptors supported or device error)
DEBUG:pyvisa:
DEBUG:pyvisa:----------
WARNING:pyvisa:User does not have permission to write to /dev/bus/usb/001/016, so the above USB INSTR device cannot be used by pyvisa; see https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/faq.html for more info.
WARNING:pyvisa:Found a USB RAW device whose serial number cannot be read. The partial VISA resource name is: USB0::1027::24593::???::0::RAW
DEBUG:asyncio:Using selector: EpollSelector
```

pyvisa_py/usb.py
```diff
@@ -292,6 +295,39 @@ def list_resources() -> List[str]:
                         "usb_interface_number": intfc,
                     },
                 )
+                logging_level = logger.getEffectiveLevel()
+                if logging_level <= logging.DEBUG:
+                    if exc_strs := traceback.format_exception(err):
```
arr-ee, line 300: what's the benefit of doing this over `logger.debug("error while reading serial number", exc_info=err)`?

I am also not sure if this amount of pre-processing is worth it. On `DEBUG` I'd prefer to have all exceptions logged in general. For higher levels, we can just add the exception message to the warning's dict maybe?

pyvisa_py/usb.py
```diff
@@ -336,7 +372,7 @@ def list_resources() -> List[str]:
                 serial = dev.serial_number
             except (NotImplementedError, ValueError, usb.USBError):
                 msg = (
-                    "Found a device whose serial number cannot be read."
+                    "Found a USB RAW device whose serial number cannot be read."
```
arr-ee, line 375: can't we run into the same permissions issue here?

pyvisa_py/usb.py
```diff
@@ -292,6 +295,39 @@ def list_resources() -> List[str]:
...
+                    missing_perms = []
+                    if not os.access(dev_path, os.O_RDONLY):
+                        missing_perms.append("read from")
+                    if not os.access(dev_path, os.O_WRONLY):
+                        missing_perms.append("write to")
+                    missing_perms_str = " or ".join(missing_perms)
+                    logger.warning(
+                        "User does not have permission to %s %s, so the above USB INSTR"
+                        " device cannot be used by pyvisa; see"
+                        " https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/faq.html"
```
arr-ee, line 326: should link to the specific anchor once docs are updated

arr-ee (changes requested):
Thank you for tackling this! Left some notes, will test the code in a bit.

Please note I am do not have authority over the code, @MatthieuDartiailh has the last word. My comments are suggestions at best.

pyvisa_py/usb.py
jllanfranchi: Only issue with that is we need to keep the two in sync; a change in the docs will necessitate a change in the code. That could be annoying for a maintainer, so I left it out. But if that's what you want, I'm happy add the anchor to the link.

pyvisa_py/usb.py
jllanfranchi: I was unaware of that kwarg to logger methods! Will change to using `exc_info`.

pyvisa_py/usb.py
```diff
@@ -292,6 +295,39 @@ def list_resources() -> List[str]:
...
+                                logger.debug(line)
+                        logger.debug("-" * len(msg))
+                elif logging_level <= logging.INFO:
+                    if exc_strs := traceback.format_exception_only(err):
+                        logger.info(
+                            "Error raised from underlying module (pyusb): %s",
+                            exc_strs[0].strip(),
+                        )
+
+                dev_path = f"/dev/bus/usb/{dev.bus:03d}/{dev.address:03d}"
```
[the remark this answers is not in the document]
jllanfranchi: As to your second point, particularly if we are going to implement this for both USB INSTR and USB RAW devices, then it certainly makes sense to move the check logic into a separate module. But I can do that anyway even just to keep the code here cleaner, omitting the gory details that are only pertinent to one OS.

As to your first comment, I thought about gating based on platform. (I also tried to figure out if there's an equivalent for FreeBSD/OS X, but failed to find that.) But I figured by the time I went through all the logic required to check that the `sys.platform` was something this check could work on, I could do it the Python-ic (duck-typing) way and simply look for the file's existence, check the file's permissions, and go from there, and that would essentially take care of "does this check work on my particular platform?" issues.

I'm certainly not married to that way, I get why one might not want to use the duck-type perspective for this check. This almost certainly only works on Linux & nowhere else, and maybe we discover a Linux where it doesn't work, and then we would just want to add to the gating condition. So, yeah, I'll put the check within a `if sys.platform.startswith("linux"): ...` check.

jllanfranchi:
Really appreciate all your helpful feedback, @arr-ee . I will push updates according to your comments ASAP, and hopefully have a first pass at adding to the docs that can be reviewed as well.

I want to re-iterate how helpful the feedback you gave is, not really being aware of all the ins-and-outs of pyvisa & pyvisa-py (like code organization in the usbutils module; wouldn't have thought of that as an outsider).

pyvisa_py/usb.py
arr-ee: I doubt that docs organisation will change drastically enough to make this a big issue, and since we're going to the trouble of linking we might as well link to the exact spot.

pyvisa_py/usb.py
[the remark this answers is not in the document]
arr-ee: > But I can do that anyway even just to keep the code here cleaner, omitting the gory details that are only pertinent to one OS.

Yup, plus it will be easier to test (that's a tricky thing though).

> But I figured by the time I went through all the logic required to check that the sys.platform was something this check could work on, I could do it the Python-ic (duck-typing) way and simply look for the file's existence, check the file's permissions, and go from there, and that would essentially take care of "does this check work on my particular platform?" issues.

This is fair! I prefer an explicit check to make the intent explicit to the readers: "this is only for linux", but my disregard for pythonic way of doing things can and should be questioned :)

> I also tried to figure out if there's an equivalent for FreeBSD/OS X, but failed to find that.

Yeah it's very system-specific, and I don't think we'd have permissions issue on macos (we'd have other ones!). I was wondering if there's a "better" way to get to the device file on linux as well, but it seems that string magic should do the trick until proven otherwise.

pyvisa_py/usb.py
MatthieuDartiailh: I agree with @arr-ee here. Please update the code accordingly.

pyvisa_py/usb.py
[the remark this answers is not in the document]
MatthieuDartiailh: Here also I think an explicit check will make the intent clearer for people looking at it.

pyvisa_py/usb.py
MatthieuDartiailh: I agree a direct link would be better. Anything that can reduce the number of obscure bug reports is something I am happy to include.

pyvisa_py/usb.py
MatthieuDartiailh: USB RAW are weird because there are supported by NI but are not part of the IVI specs. If we can provide more information we should though.

MatthieuDartiailh (changes requested):
Looks good. Once the minor requested changes are in I will merge.

MatthieuDartiailh:
@jllanfranchi any idea when you will be able to address the comments ?

MatthieuDartiailh:
Friendly ping @jllanfranchi

pyvisa_py/usb.py
DavidLutton: I observed in pyvisa issue [Linux host and getting permissions to talk to a USBTMC device](https://github.com/pyvisa/pyvisa/issues/609#issuecomment-925396821) 
USBTMC devices show up as both /dev/usbtmcN and /dev/bus/usb/NNN/NNN
Both instances need to have suitable permissions for pyvisa with pyvisa-py to be able to communicate

If only the /dev/usbtmcN has permissions then you get `WARNING Found a device whose serial number cannot be read`

MatthieuDartiailh:
I took the liberty to rebase and address the log and the platform gating. @arr-ee we have no entry in the FAQ for the permissions but we do have one in the installation. Would you have time to look at how to improve it ? in particular it may make sense to include @DavidLutton latest comment somewhere.

MatthieuDartiailh:
Friendly ping @arr-ee

MatthieuDartiailh:
Friendly ping @arr-ee

jllanfranchi:
@MatthieuDartiailh I apologize I completely dropped the ball on this, and my job took me largely away from software development so I haven't been active here. But, yes, in hindsight this all addressed what my original issues were. Thank you for picking up the ball I dropped, Matthieu.

Issue #436, opened 2024-06-11
Communication issue with Siglent SDS2352XE using pyvisa-py


JaminMartin:
Instrument details
------------------
* Siglent SDS2352XE
* USB
* https://www.siglent.eu/product/1139322/siglent-sds2352x-e-350mhz-dual-channel-oscilloscope

Output of `pyvisa-info`
```
Machine Details:
   Platform ID:    Linux-6.5.0-35-generic-x86_64-with-glibc2.35
   Processor:      x86_64

Python:
   Implementation: CPython
   Executable:     /home/jamin/miniforge3/envs/spcs_infra/bin/python
   Version:        3.12.3
   Compiler:       GCC 12.3.0
   Architecture:   ('x86', 64)
   Build:          Apr 15 2024 18:38:13 (#main)
   Unicode:        UCS4

PyVISA Version: 1.14.1

Backends:
   ivi:
      Version: 1.14.1 (bundled with PyVISA)
      #1: /usr/lib/x86_64-linux-gnu/libvisa.so.24.0.0:
         found by: auto
         architecture:
            ('x86', 64)
         Vendor: National Instruments
         Impl. Version: 25165824
         Spec. Version: 7340032
   py:
      Version: 0.7.2
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR: Available via PyUSB (1.2.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.2.1). Backend: libusb1
      TCPIP INSTR: Available 
         Resource discovery:
         - VXI-11: partial (psutil not installed)
         - hislip: disabled (zeroconf not installed)
      TCPIP SOCKET: Available 
      VICP INSTR:
         Please install PyVICP to use this resource type.
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
      GPIB INTFC:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
```
Hello, I am having communication issues with an oscilloscope only using the pyvisa-py back-end. The issue seems to be similar to the issues presented [here](https://github.com/pyvisa/pyvisa/issues/158#issuecomment-2161513955) 

This is a minimal configuration: 

import pyvisa

[...]

Issue #318, opened 2022-02-20
[COM] Communication issue with XXX using XXX


WillMatthews:
Instrument details
------------------
* Model: Siglent SDS 1104X-E
* Communication: USB (usbtmc I if I am sane?)
* Link to the documentation: [https://siglentna.com/wp-content/uploads/dlm_uploads/2017/10/ProgrammingGuide_forSDS-1-1.pdf](https://siglentna.com/wp-content/uploads/dlm_uploads/2017/10/ProgrammingGuide_forSDS-1-1.pdf)

Output of `pyvisa-info`
-----------------------
```
Machine Details:
   Platform ID:    Linux-5.4.0-99-generic-x86_64-with-Ubuntu-18.04-bionic
   Processor:      x86_64

Python:
   Implementation: CPython
   Executable:     /usr/bin/python3
   Version:        3.6.9
   Compiler:       GCC 8.4.0
   Bits:           64bit
   Build:          Dec  8 2021 21:08:43 (#default)
   Unicode:        UCS4

PyVISA Version: 1.11.3

Backends:
   ivi:
      Version: 1.11.3 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.5.2
      ASRL INSTR: Available via PySerial (3.5)
      USB INSTR: Available via PyUSB (1.1.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.1.1). Backend: libusb1
      TCPIP INSTR: Available 
      TCPIP SOCKET: Available 
      GPIB INSTR:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of funcionality.
         No module named 'gpib'
```

The problem...
------------------

I got this traceback, which I've seen mentioned around with no good fix for..

`Traceback (most recent call last):
  File "./test-visa.py", line 172, in <module>
    data = scope.query_binary_values("C1:WF? DAT2", datatype='b')
  File "/home/will/.local/lib/python3.6/site-packages/pyvisa/resources/messagebased.py", line 754, in query_binary_values
    chunk_size,
  File "/home/will/.local/lib/python3.6/site-packages/pyvisa/resources/messagebased.py", line 602, in read_binary_values
    self.read_bytes(expected_length - len(block), chunk_size=chunk_size)
[...]

Issue #311, opened 2021-12-07
pyvisa-py backend problem writing data files via tcp containing over 450 points (900 bytes) to a Siglent SDG2042x AWG. 


dirtcooker:
when writing fewer than about 450 points via the following two lines of code, the data is received properly on the awg. More than that and the bytes written by the 2nd line end up in the data file on the awg, rather than being executed as a separate command. The full script is attached. Note that the first line of code below creates a file on the awg containing the waveform data. The 2nd line tells the awg to load it.

dev.write("C1:WVDT WVNM,wave2,FREQ,100.0,AMPL,15.0,OFST,0.0,PHASE,0.0,LENGTH,%d,WAVEDATA,%s"%(nbytes,data),encoding='latin1') 
dev.write("C1:ARWV NAME,wave2")

The script works properly on a win10 machine using the NI backend. Thus I suspect this is a bug in the pyvisa-py backend. On a linux mint machine (same behavior on v19.2 and 20.2) using pyvisa-py it exhibits the erroneous behavior. I have copied the file from the awg back to the pc and opened it in a hex editor. Attached is a pic of the end of the data file, showing what happened. 
a work-around is to insert the following line in between the above two statements, which adds an additional 2 bytes to the data file, but seems to then properly close the data file, so that the next write is interpreted properly by the awg.
dev.write('\x00\x80',encoding='latin1')
--------------
Steps to reproduce the behavior: run the attached script.

**Output of `pyvisa-info`**

Python:
   Implementation: CPython
   Executable:     /usr/bin/python3
   Version:        3.6.9
   Compiler:       GCC 8.4.0
   Bits:           64bit
   Build:          Jan 26 2021 15:33:00 (#default)
   Unicode:        UCS4
[sdgarb.py.zip](https://github.com/pyvisa/pyvisa-py/files/7664615/sdgarb.py.zip)
![corrupted wave file](https://user-images.githubusercontent.com/55506744/144946792-d1ad45b5-cfe0-4b73-bc30-1d09f538b181.png)

PyVISA Version: 1.11.3

Backends:
   ivi:
      Version: 1.11.3 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.5.2

PR #449, opened 2024-09-10
Fixed bug in USBTMC code which caused issues when packets were longer than max usb packet size

hcoffin-vecatom:
Following sec. 3.3 of "Universal Serial Bus Test and Measurement Class Specification (USBTMC)" Rev 1.0, updated the read function of USBTMC in protocols/usbtmc.py . Read now reads from usb until a "short packet" has been read (see specification), and only expects a header on the first packet received. In usb.py, added separate read function to USBInstrSession which is nearly identical to USBSessions.read, but has "lambda current: False" instead of "lambda current: True", to allow for usbtmc device to return multiple packets before a termination character and to remove trailing alignment bytes which might be sent by device.

### DIFF
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -347,7 +347,8 @@ class USBInstrSession(USBSession):
         return self._read(
             _usb_reader,
             count,
-            lambda current: False,  # USBTMC can return partial message (i.e., before the term_char) or have trailing zeros
+            lambda current: False,  # USBTMC can return partial message (i.e.,
+            # before the term_char) or have trailing zeros
             supress_end_en,
             term_char,
             term_char_en,
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -157,11 +157,6 @@ class USBSession(Session):
 
         supress_end_en, _ = self.get_attribute(ResourceAttribute.suppress_end_enabled)
 
-        if supress_end_en:
-            raise ValueError(
-                "VI_ATTR_SUPPRESS_END_EN == True is currently unsupported by pyvisa-py"
-            )
-
         term_char, _ = self.get_attribute(ResourceAttribute.termchar)
         term_char_en, _ = self.get_attribute(ResourceAttribute.termchar_enabled)
 
@@ -305,6 +300,7 @@ class USBInstrSession(USBSession):
                 }
             )
         return out
+
 
     def read(self, count: int) -> Tuple[bytes, StatusCode]:
         """Reads data from device or interface synchronously.
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -301,55 +301,6 @@ class USBInstrSession(USBSession):
             )
         return out
 
-    def read(self, count: int) -> Tuple[bytes, StatusCode]:
-        """Reads data from device or interface synchronously.
-
-        Corresponds to viRead function of the VISA library.
-
-        Parameters
-        -----------
-        count : int
-            Number of bytes to be read.
-
-        Returns
-        -------
-        bytes
-            Data read from the device
-        StatusCode
-            Return value of the library call.
-
-        """
-
-        def _usb_reader():
-            """Data reader identifying usb timeout exception."""
-            try:
-                return self.interface.read(count)
-            except usb.USBError as exc:
-                if exc.errno in (errno.ETIMEDOUT, -errno.ETIMEDOUT):
-                    raise USBTimeoutException()
-                raise
-
-        supress_end_en, _ = self.get_attribute(ResourceAttribute.suppress_end_enabled)
-
-        if supress_end_en:
-            raise ValueError(
-                "VI_ATTR_SUPPRESS_END_EN == True is currently unsupported by pyvisa-py"
-            )
-
-        term_char, _ = self.get_attribute(ResourceAttribute.termchar)
-        term_char_en, _ = self.get_attribute(ResourceAttribute.termchar_enabled)
-
-        return self._read(
-            _usb_reader,
-            count,
-            lambda current: False,  # USBTMC can return partial message (i.e.,
-            # before the term_char) or have trailing zeros
-            supress_end_en,
-            term_char,
-            term_char_en,
-            USBTimeoutException,
-        )
-
 
 @Session.register(constants.InterfaceType.usb, "RAW")
 class USBRawSession(USBSession):
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -305,6 +305,55 @@ class USBInstrSession(USBSession):
                 }
             )
         return out
+    
+    def read(self, count: int) -> Tuple[bytes, StatusCode]:
+        """Reads data from device or interface synchronously.
+
+        Corresponds to viRead function of the VISA library.
+
+        Parameters
+        -----------
+        count : int
+            Number of bytes to be read.
+
+        Returns
+        -------
+        bytes
+            Data read from the device
+        StatusCode
+            Return value of the library call.
+
+        """
+
+        def _usb_reader():
+            """Data reader identifying usb timeout exception."""
+            try:
+                return self.interface.read(count)
+            except usb.USBError as exc:
+                if exc.errno in (errno.ETIMEDOUT, -errno.ETIMEDOUT):
+                    raise USBTimeoutException()
+                raise
+
+        supress_end_en, _ = self.get_attribute(ResourceAttribute.suppress_end_enabled)
+
+        if supress_end_en:
+            raise ValueError(
+                "VI_ATTR_SUPPRESS_END_EN == True is currently unsupported by pyvisa-py"
+            )
+
+        term_char, _ = self.get_attribute(ResourceAttribute.termchar)
+        term_char_en, _ = self.get_attribute(ResourceAttribute.termchar_enabled)
+
+        return self._read(
+            _usb_reader,
+            count,
+            lambda current: False,  # USBTMC can return partial message (i.e., before the term_char) or have trailing zeros
+            supress_end_en,
+            term_char,
+            term_char_en,
+            USBTimeoutException,
+        )
+
 
 
 @Session.register(constants.InterfaceType.usb, "RAW")
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -301,7 +301,6 @@ class USBInstrSession(USBSession):
             )
         return out
 
-
     def read(self, count: int) -> Tuple[bytes, StatusCode]:
         """Reads data from device or interface synchronously.
 
--- a/pyvisa_py/usb.py
+++ b/pyvisa_py/usb.py
@@ -305,7 +305,7 @@ class USBInstrSession(USBSession):
                 }
             )
         return out
-    
+
     def read(self, count: int) -> Tuple[bytes, StatusCode]:
         """Reads data from device or interface synchronously.
 
@@ -354,7 +354,6 @@ class USBInstrSession(USBSession):
             term_char_en,
             USBTimeoutException,
         )
-
 
 
 @Session.register(constants.InterfaceType.usb, "RAW")
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,10 @@ PyVISA-py Changelog
 ------------------
 
 - add read_stb method for TCPIP HiSLIP client PR #429
+- fix usbtmc implementation to respect section 3.3 of the spec PR #449
+  Read now reads from usb until a "short packet" has been read
+  (see specification), and only expects a header on the first packet received.
+- add support for VI_ATTR_SUPPRESS_END_EN for USB resources PR #449
 
 0.7.2 (07/03/2024)
 ------------------
--- a/pyvisa_py/protocols/usbtmc.py
+++ b/pyvisa_py/protocols/usbtmc.py
@@ -481,9 +481,9 @@ class USBTMC(USBRaw):
                 response = BulkInMessage.from_bytes(resp)
                 received.extend(response.data)
                 while len(resp) == self.usb_recv_ep.wMaxPacketSize:
-                    # USBTMC Section 3.3 specifies that the first usb packet 
-                    # must contain the header. the remaining packets do not need 
-                    # the header the message is finished when a "short packet" 
+                    # USBTMC Section 3.3 specifies that the first usb packet
+                    # must contain the header. the remaining packets do not need
+                    # the header the message is finished when a "short packet"
                     # is sent (one whose length is less than wMaxPacketSize)
                     resp = raw_read(recv_chunk + header_size + max_padding)
                     received.extend(resp)
--- a/pyvisa_py/protocols/usbtmc.py
+++ b/pyvisa_py/protocols/usbtmc.py
@@ -479,12 +479,18 @@ class USBTMC(USBRaw):
             try:
                 resp = raw_read(recv_chunk + header_size + max_padding)
                 response = BulkInMessage.from_bytes(resp)
+                received.extend(response.data)
+                while len(resp) == self.usb_recv_ep.wMaxPacketSize:
+                    # USBTMC Section 3.3 specifies that the first usb packet must contain the header
+                    # the remaining packets do not need the header
+                    # the message is finished when a "short packet" is sent (one whose length is less than wMaxPacketSize)
+                    resp = raw_read(recv_chunk + header_size + max_padding)
+                    received.extend(resp)
             except (usb.core.USBError, ValueError):
                 # Abort failed Bulk-IN operation.
                 self._abort_bulk_in(self._btag)
                 raise
 
-            received.extend(response.data)
 
             # Detect EOM only when device sends all expected bytes.
             if len(response.data) >= response.transfer_size:
--- a/pyvisa_py/protocols/usbtmc.py
+++ b/pyvisa_py/protocols/usbtmc.py
@@ -481,16 +481,16 @@ class USBTMC(USBRaw):
                 response = BulkInMessage.from_bytes(resp)
                 received.extend(response.data)
                 while len(resp) == self.usb_recv_ep.wMaxPacketSize:
-                    # USBTMC Section 3.3 specifies that the first usb packet must contain the header
-                    # the remaining packets do not need the header
-                    # the message is finished when a "short packet" is sent (one whose length is less than wMaxPacketSize)
+                    # USBTMC Section 3.3 specifies that the first usb packet 
+                    # must contain the header. the remaining packets do not need 
+                    # the header the message is finished when a "short packet" 
+                    # is sent (one whose length is less than wMaxPacketSize)
                     resp = raw_read(recv_chunk + header_size + max_padding)
                     received.extend(resp)
             except (usb.core.USBError, ValueError):
                 # Abort failed Bulk-IN operation.
                 self._abort_bulk_in(self._btag)
                 raise
-
 
             # Detect EOM only when device sends all expected bytes.
             if len(response.data) >= response.transfer_size:
--- a/pyvisa_py/sessions.py
+++ b/pyvisa_py/sessions.py
@@ -796,10 +796,9 @@ class Session(metaclass=abc.ABCMeta):
             if current:
                 out.extend(current)
                 end_indicator_received = end_indicator_checker(current)
-                if end_indicator_received:
-                    if not suppress_end_en:
-                        # RULE 6.1.1
-                        return bytes(out), StatusCode.success
+                if end_indicator_received and not suppress_end_en:
+                    # RULE 6.1.1
+                    return bytes(out), StatusCode.success
                 else:
                     if termination_char_en and (term_char in current):
                         # RULE 6.1.2


arr-ee:
Hi — thank you for the PR!

Would you mind providing some context regarding what prompted you to make the change? Any specific instrument/command pattern that were having issues and are fixed by this? This information might help with revisiting existing USBTMC-related issues.

Thanks!

hcoffin-vecatom:
> Hi — thank you for the PR!
> 
> Would you mind providing some context regarding what prompted you to make the change? Any specific instrument/command pattern that were having issues and are fixed by this? This information might help with revisiting existing USBTMC-related issues.
> 
> Thanks!

I am working with a Siglent SDG1062X function generator. I tried using querying the waveform parameters using 'CH1:BSWV?' (see https://siglentna.com/USA_website_2014/Documents/Program_Material/SDG_ProgrammingGuide_PG_E03B.pdf section 3.4), but I was getting issues where I would only get a partial response back (first 52 bytes, which is one usb packet without the 12 byte BulkInMessage header), and future queries would return garbage (the remaining packets partially misinterpreted as part of the BulkInMessage header information). The code in this PR fixes the issue for me, but I haven't been able to test it on other devices, so I'm not sure if part of this is a quirk of this particular device.

arr-ee:
Amazing, thank you so much!

This PR might be related to a bunch of existing issues around USBTMC connectivity with Siglent and Rigol instruments (that are known to be... quirky): off the top of the issues list, #414, #436, #318, and #311 seem like they might be caused by the same problem.

Once again, thank you for taking the time to dig into the issue and open a PR.

hcoffin-vecatom:
> Amazing, thank you so much!
> 
> This PR might be related to a bunch of existing issues around USBTMC connectivity with Siglent and Rigol instruments (that are known to be... quirky): off the top of the issues list, #414, #436, #318, and #311 seem like they might be caused by the same problem.
> 
> Once again, thank you for taking the time to dig into the issue and open a PR.

Happy to help! Took a look at the issues you mentioned. #414 and #436 look like the same issue I was having. #318 could be, but has a slightly different symptom than what I was seeing; in the USBTMC spec they say that the "short message" can be 0 bytes, which sounds kind of like the symptom they describe. #311 definitely won't be fixed by this PR since I didn't change any of the write code. I didn't look into the spec for bulkout at all so I'm not sure if it has the same type of issue.

pyvisa_py/usb.py
```diff
@@ -306,6 +306,55 @@ def list_resources() -> List[str]:
...
+                "VI_ATTR_SUPPRESS_END_EN == True is currently unsupported by pyvisa-py"
+            )
+
+        term_char, _ = self.get_attribute(ResourceAttribute.termchar)
+        term_char_en, _ = self.get_attribute(ResourceAttribute.termchar_enabled)
+
+        return self._read(
+            _usb_reader,
+            count,
+            lambda current: False,  # USBTMC can return partial message (i.e.,
```
MatthieuDartiailh, line 350: Here the end indicator checker does not care about the termination character, only about the fact that the underlying hardware signaled it is done transmitting a message. For GPIB, this is done by checking the state of hardware line, for USB it is always trivially true since by construction a BulkIn message is complete when received. Using this kind of notification is more efficient than checking for term chars.

The ``_read`` implementation takes care of ensuring we read till a term char if we were requested to do so. Actually it is at this layer that support for ``suppress_end`` is implemented so I would appreciate if while removing this implementation, you could also get rid of the error forbidding the use of ``suppress_end``.

MatthieuDartiailh (changes requested):
Thanks your contribution ! USBTMC is complicated and all fix improving support are more than welcome.

I left a comment that I think is worth addressing here. Can you have a look @hcoffin-vecatom ? 

In addition to that if you can add a releasenotes entry it would be great.

hcoffin-vecatom:
> Thanks your contribution ! USBTMC is complicated and all fix improving support are more than welcome.
> 
> I left a comment that I think is worth addressing here. Can you have a look @hcoffin-vecatom ?
> 
> In addition to that if you can add a releasenotes entry it would be great.


I think I have made the relevant changes. I'm not sure if I did the releasenotes correctly as I haven't done that before, so feel free to change it if desired.

hcoffin-vecatom:
Also here is the code I used to test it in case someone else needs it:
```
import pyvisa

res = pyvisa.ResourceManager('@py').open_resource('USB0::62700::4355::SDG1XDDC801438::0::INSTR')
res.set_visa_attribute(pyvisa.constants.VI_ATTR_SUPPRESS_END_EN, True)
res.read_termination = '\n'
res.write_termination = '\n'

with res:
    res.write('C1:BSWV?')
    print(res.read_raw())
```

With 'res.set_visa_attribute(pyvisa.constants.VI_ATTR_SUPPRESS_END_EN, True)' it outputs
```
b'C1:BSWV WVTP,SINE,FRQ,123HZ,PERI,0.00813008S,AMP,0.05V,AMPVRMS,0.017675Vrms,AMPDBM,0dBm,MAX_OUTPUT_AMP,20V,OFST,-0.2V,HLEV,-0.175V,LLEV,-0.225V,PHSE,78\n'
```

With 'res.set_visa_attribute(pyvisa.constants.VI_ATTR_SUPPRESS_END_EN, False)' it outputs
```
b'C1:BSWV WVTP,SINE,FRQ,123HZ,PERI,0.00813008S,AMP,0.05V,AMPVRMS,0.017675Vrms,AMPDBM,0dBm,MAX_OUTPUT_AMP,20V,OFST,-0.2V,HLEV,-0.175V,LLEV,-0.225V,PHSE,78\n\x00\x00\x00\x00\x00\x00\x00\x00'
```

MatthieuDartiailh:
Wow thanks for doing the fixes so fast !!!

For the suppress_end option, can you run your test with NI or Keysight VISA to see if the output is the same ?

hcoffin-vecatom:
> Wow thanks for doing the fixes so fast !!!
> 
> For the suppress_end option, can you run your test with NI or Keysight VISA to see if the output is the same ?

Unfortunately, I have not been unable to correctly install either of those libraries on my computer (running Linux Mint w/ Ubuntu 21.2). I tried installing NI-VISA before looking into changing the code in pyvisa-py and was unable to get it to even see any device let alone query them.

I just tried installing Keysight's version, but I'm mostly getting errors (to be fair to Keysight, they say my version of Linux isn't supported during the installation).
Every once in a while, if keep unplugging and replugging the device with suppress_end set to False, I get
```
b'C1:BSWV WVTP,SINE,FRQ,1000HZ,PERI,0.001S,AMP,4V,AMPV'
Segmentation fault (core dumped)
```
which (besides the seg fault) is what I got with pyvisa-py backend before any of the updates in this pull request. Trying to run the test script again without replugging the device results in.
```
pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
```

If I run it with suppress_end set to True, I always get
```
pyvisa.errors.VisaIOError: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
```

So I would not trust that suppress_end is working correctly unless someone else can verify.

MatthieuDartiailh:
@arr-ee would be able to test the suppress_end behavior ?

arr-ee:
I should be able to in the next couple of days, will update soon

arr-ee:
Testing with my SDG2122X:

### Test code:

```python
import sys
import pyvisa

pyvisa.log_to_screen()

i = rm.open_resource('USB0::62700::60984::SDG2XCA4162836::0::INSTR')
i.set_visa_attribute(pyvisa.constants.VI_ATTR_SUPPRESS_END_EN, sys.argv[1] == '1')

print(f"IDN: {i.query('*IDN?')}")
i.write('C1:BSWV?')
print(f"C1:BSWV?: {i.read_raw()}")
```

### pyvisa 1.14.1, pyvisa-py 0.7.2:

Repeated executions of the script lead to different results, with previous responses showing up in place of up to date ones, e.g.:

```
$ python test.py 0
...
USB0::62700::60984::SDG2XCA4162836::0::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_base/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py:116: UserWarning: Unexpected MsgID format. Consider updating the device's firmware. See https://github.com/pyvisa/pyvisa-py/issues/20Expected message id was 2, got 77.
  warnings.warn(
IDN: C1:BSWV WVTP,SINE,FRQ,55000000HZ,PERI,1.81818e-08S,AMP,1
2024-09-20 16:22:28,308 - pyvisa - DEBUG - USB0::62700::60984::SDG2XCA4162836::0::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)
/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_base/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py:116: UserWarning: Unexpected MsgID format. Consider updating the device's firmware. See https://github.com/pyvisa/pyvisa-py/issues/20Expected message id was 2, got 46.
  warnings.warn(
C1:BSWV?: b'Siglent Technologies,SDG2122X,SDG2XCA4162836,2.01.01.35R'
2024-09-20 16:22:28,309 - pyvisa - DEBUG - Closing ResourceManager (session: 4947628)
```

`pyvisa.constants.VI_ATTR_SUPPRESS_END_EN` is not supported by the released version.

### pyvisa 1.14.1, pyvisa-py 0.7.3.dev26+gc632e68:

Same code hangs on the first read, not reacting to Ctrl-C. When process is killed, attempts to write to the device timeout, and even physical controls do not react to anything (I suspected it to be "remote" mode, but hard-reboot and running the script using the released version does not lead to UI lock-up). Adding read/write termination does not change the behaviour.

I also tried running the same code with R&S VISA, and both through pyvisa and through the bundled tester utility I've been getting `exception while reading: VI_ERROR_IO (-1073807298): Could not perform operation because of I/O error.` when reading `C1:BSWV?` response.

Attached are packet captures for released code (before), new code running after released (after), new code running after instrument reboot (after_clean), and rsvisa (rsvisa)

[pyvisapy_449_captures.zip](https://github.com/user-attachments/files/17078048/pyvisapy_449_captures.zip)

arr-ee:
However, I do see progress when running a slightly changed test code against SDS2104X Plus:

Code change:

```python
print(f"IDN: {i.query('*IDN?')}")
i.write(':WAVEFORM:SOURCE C1')
i.write(':WAVEFORM:DATA?')
print(f":WAVEFORM:DATA?: {i.read_raw()}")
```

Released code gets the same issues with old results, but that also leads to 

```
...
  File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_base/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 481, in read
    response = BulkInMessage.from_bytes(resp)
  File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_base/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 114, in from_bytes
    msgid, btag, btaginverse = struct.unpack_from("BBBx", data)
struct.error: unpack_from requires a buffer of at least 4 bytes for unpacking 4 bytes at offset 0 (actual buffer size is 0)
```

errors. R&S VISA just throws I/O errors on raw read.

Updated code (after several tried to get rid of old replies in the instrument buffer) works every time. `suppress_end` also appears to be working since I do sometimes get implausibly short results when it is enabled because response can include `\n` characters.

arr-ee:
According to a quick glance at the packet capture, version from the PR attempts to read repeatedly (?), and responses from the device get more and more jumbled until it hangs. I have not looked into USB packet internals much before, so it might take me a minute. 

@hcoffin-vecatom let me know if there is any info you might need in the meantime, I have aforementioned Siglent devices, some supposedly better behaving ones (R&S/HPAK), and some potentially broken-in-different-ways (hi GW Instek and possibly Tenma) that I'll try to test against.

Another thing I can eventually try is updating the firmware on the SDG — it's several minor versions behind, although it's not _old_ old; and testing it against a windows box with Siglent's drivers to get a "reference implementation" (assuming it works).

hcoffin-vecatom:
Took me a bit to parse those log files, but it looks like your SDG2122X has some even weirder behavior than my SDG1062X.
When I do '*IDN?', I get the following packets
```
b'\x02\x02\xfd\x004\x00\x00\x00\x00\x00\x00\x00Siglent Technologies,SDG1062X,SDG1XDDC801438,1.01.01'
b'.33R'
b'\x02\x03\xfc\x00\x06\x00\x00\x00\x01\x00\x00\x008\n\x00\x00\x00\x00\x00\x00'
```
Notice the x01 on the last packet. That represents that this usbtmc message (which could be multiple usb packets) should be the last (the eom logic in usbtmc.py).

When you do the same query on your device (looking at the after_clean capture you sent), you never receive an eom (in picture marked below, I have underlined where that should be in red). The device sends a new line character (underlined in green), which I guess is meant to tell the computer to stop reading more packets, but that is not how usbtmc is specified. It then proceeds to just keep copy and pasting its id over and over.
image

To be honest, I have no idea how to handle that in a reasonable way. I guess you could set a relatively short timeout and just get rid of all the extra copies it sends, but that seems quite hacky. Only other option I can see is to parse out termination characters in the USBTMC.read function, which again seems wrong.

arr-ee:
Fascinating.

@hcoffin-vecatom thank you for looking into the captures right away.

I’ll try to run more tests in the next couple of days to hopefully get a better understanding of what’s happening, but ultimately I think this PR should be OK to merge since my instrument’s behaviour appears to be broken-broken.

If no one minds I suggest postponing merge until early next week just in case any of further tests point to an explanation/workaround.

hcoffin-vecatom:
> Fascinating.
> 
> @hcoffin-vecatom thank you for looking into the captures right away.
> 
> I’ll try to run more tests in the next couple of days to hopefully get a better understanding of what’s happening, but ultimately I think this PR should be OK to merge since my instrument’s behaviour appears to be broken-broken.
> 
> If no one minds I suggest postponing merge until early next week just in case any of further tests point to an explanation/workaround.

Thought of one more small thing to try over the weekend. I'm guessing you have already tried this, but if you haven't, can you try setting the termination character to '\n' before reading/writing to SDG2122X. It looks like, at least for writing to the device, you are terminating with '\r\n' in that log. The SDG programming manual specifies '\n' as the termination, so I'm not sure what to expect if it isn't.
image

arr-ee:
Yes, I have noticed that you were changing termination settings, and tried that as well with no change in behaviour IIRC. I will try it again to make sure I have not missed a combination of factors that actually works.

arr-ee:
Thank you for your patience, took me a minute.

I only got to run linux-based tests since windows machine is out of reach for this instrument's USB right now, but tl;dr PR is fine, Siglent just being Siglent. Firmware update from `2.01.01.35R` (2019 vintage) to latest fixed hanging reads.

Details:

1. Changing read/write termination does not change the behaviour, the first read still hangs:
    ```
    <powering the DUT off during hang produces the following>
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 480, in read
        resp = raw_read(recv_chunk + header_size + max_padding)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 295, in read
        data = self.usb_recv_ep.read(size, self.timeout).tobytes()
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/core.py", line 423, in read
        return self.device.read(self, size_or_buffer, timeout)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/core.py", line 1029, in read
        ret = fn(
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/backend/libusb1.py", line 846, in bulk_read
        return self.__read(self.lib.libusb_bulk_transfer,
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/backend/libusb1.py", line 954, in __read
        _check(retval)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/backend/libusb1.py", line 604, in _check
        raise USBError(_strerror(ret), ret, _libusb_errno[ret])
    usb.core.USBError: [Errno 5] Input/Output Error
    
    During handling of the above exception, another exception occurred:
    
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/test.py", line 33, in <module>
        logger.info(f"IDN: {i.query('*IDN?')}")
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa/resources/messagebased.py", line 647, in query
        return self.read()
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa/resources/messagebased.py", line 485, in read
        message = self._read_raw().decode(enco)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa/resources/messagebased.py", line 441, in _read_raw
        chunk, status = self.visalib.read(self.session, size)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/highlevel.py", line 513, in read
        data, status_code = self.sessions[session].read(count)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/usb.py", line 163, in read
        return self._read(
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/sessions.py", line 792, in _read
        current = reader()
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/usb.py", line 152, in _usb_reader
        return self.interface.read(count)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 492, in read
        self._abort_bulk_in(self._btag)
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/pyvisa_py/protocols/usbtmc.py", line 388, in _abort_bulk_in
        data = self.usb_dev.ctrl_transfer(
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/core.py", line 1082, in ctrl_transfer
        ret = self._ctx.backend.ctrl_transfer(
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/backend/libusb1.py", line 893, in ctrl_transfer
        ret = _check(self.lib.libusb_control_transfer(
      File "/home/lab/projects/pyvisa-repro/449-usbtmc/.venv_fixed/lib/python3.10/site-packages/usb/backend/libusb1.py", line 604, in _check
        raise USBError(_strerror(ret), ret, _libusb_errno[ret])
    usb.core.USBError: [Errno 5] Input/Output Error
    ```
1. kernel `usbtmc` driver (`6.8.0-45-generic`, Ubuntu 24.04.1 LTS) behaves exactly similarly to R&S VISA: incomplete first read, nothing on second read:
    ```
    $ echo "*IDN?" > /dev/usbtmc0 && cat /dev/usbtmc0
    Siglent Technologies,SDG2122X,SDG2XCA4162836,2.01.01<no newline, not full fw version>
    dmesg:
    [1148895.281910] usbtmc 1-1.4.1:1.0: Device sent too small first packet: 8 < 12
    $ echo "C1:BSWV?" > /dev/usbtmc0 && cat /dev/usbtmc0
    <no output>
    dmesg:
    [1148947.020224] usbtmc 1-1.4.1:1.0: Device sent reply with wrong bTag: 3 != 5
    <(also seen Device sent reply with wrong bTag: 5 != 8)>
    ```
    Note that 6.8.0 does not include https://github.com/dpenkler/linux-usbtmc/pull/9; however, that logic has been added to `pyvisa-py` (https://github.com/pyvisa/pyvisa-py/pull/417), so it might be worth it to test the experimental kernel driver to see if it exhibits issues similar to what's being fixed here.
1. Finally, `SUPPRESS_END_EN`:
    1. Disabled: identical results for default `read_termination=None` and `read_termination='\n'`: bunch of zero bytes after `\n`.
    1. Enabled: 
        - `read_termination=None`: all reads timeout
        - `read_termination='\n'`: response string ends with `\n`.
    This appears to be the desired behaviour I think?

Apologies for not getting to the bottom of this, but at least it appears to be safe(r) to merge now.

MatthieuDartiailh:
I updated the README and will merge as soon as the CIs are green.

MatthieuDartiailh:
Thanks @hcoffin-vecatom for the patch and @arr-ee for testing !!!

arr-ee:
Huge shout to @hcoffin-vecatom for the fix and explanations during testing!

PR #452, opened 2024-09-18
readme: deduplicate requirement version specification

henriquesimoes:
Specifying the dependency versions in the README file creates an extra maintenance burden not worth the effort. In fact, it has been left outdated more than once leading to misinformation. Instead, provide the rough information in the documentation and point users to the source of truth for details.

Python 3 is still listed explicitly since it won't be bumped every now and then, and provides a useful compatibility information for users. Unfortunately, PyVISA cannot be dealt with in a similar vein.

While here, also use code block syntax for `pip install` command.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,7 +78,7 @@ Optionally:
 Installation
 --------------
 
-Using pip:
+Using pip::
 
     $ pip install pyvisa-py
 
--- a/README.rst
+++ b/README.rst
@@ -61,8 +61,8 @@ controlling:
 Requirements
 ------------
 
-- Python (tested with 3.6+)
-- PyVISA 1.11+
+- Python 3
+- PyVISA
 
 Optionally:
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -74,11 +74,13 @@ Optionally:
 - zeroconf (for HiSLIP and VICP devices discovery)
 - pyvicp (to enable the Teledyne LeCroy proprietary VICP protocol)
 
+Please refer to `pyproject.toml <./pyproject.toml>`_ for the specific version
+requirements.
```
henriquesimoes, line 78: I'm not sure if this approach I followed works okay for PyPI page, though. GitHub properly redirects to the repository file, but I cannot see how PyPI would do the same.

Let me know if you think that's an issue.

MatthieuDartiailh: You can use `twine check` on a sdist to check if the README is admissible for PyPI. I will do it next week if you do not beat me to it.

MatthieuDartiailh:
Thanks it looks good. Will merge once I confirm that PyPI will accept the README.

MatthieuDartiailh:
I could confirm the readme is fine for PyPI. So I will merge.

README.rst
henriquesimoes: Nice. I didn't know about `twine check`. Thanks for checking this out.

PR #459, opened 2024-10-11
fix link to VISA spec

kbroch-rivosinc:
existing link had 404, looks like website was updated to https and spec download page reworked


MatthieuDartiailh:
Thanks

PR #466, opened 2024-11-16
Write long data to USBTMC correctly

hzwkxbksjoqj:
Previously the variable holding the data would get overwritten by mistake when it could not fit in a single packet.

- [ ] Executed ``black . && isort -c . && flake8`` with no errors  I did not figure out how to achieve it.  I did not change the imports at the top of the file.  Isort thinks the line is too long and tries to break it into multiple lines.  Then black thinks the line is too short and changes it back.


arr-ee:
Can you please explain the mechanism by which this change fixes the issue, and provide a repro case showcasing the difference in behaviour?

hzwkxbksjoqj:
I have a Tektronix TDS580D oscilloscope and a [Xyphro GPIB adapter](https://github.com/xyphro/UsbGpib).  The adapter talks to the computer as a USBTMC device.  I was trying to copy a file to the device.  Below is the code I used.  Without this patch, the program hanged indefinitely at the `write_raw` line.  The file was 100 bytes long.  A USB packet contained 64 bytes according to `handle.visalib.sessions[handle.session].interface.usb_send_ep.wMaxPacketSize`.

```Python
import pyvisa

resources = pyvisa.ResourceManager()
devices = resources.list_resources()
assert len(devices) == 1
handle = resources.open_resource(devices[0])
handle.query("*IDN?")
with open("file", "rb") as fh:
    content = fh.read()
    size = str(len(content))
    handle.write_raw(
        f'FILESYSTEM:WRITEFILE "hd0:/FILE", #{len(size)}{size}'.encode(
            "us-ascii"
        )
        + content
    )
```

Issue #468, opened 2024-11-18
solution to fix usbtmc frame partitioning issues on Linux.


jyfrau:
Hello,

I am sharing with you a solution to fix the usbtmc frame partitioning issues that are not handled properly under Linux.

I am using pyvisa and pyvisa-py to communicate with Siglent devices (SSA3032X and SVA1032X). On Linux it does not work. I get the message "Unexpected MsgID format. Consider updating the device's firmware. See https://github.com/pyvisa/pyvisa-py/issues/20".

The problem comes from the fact that usbtmc frames are split into 64-byte packets. A problem that does not appear under Windows, nor on a Linux virtual machine on a Windows host (which therefore goes through the Windows driver...).

Although this partitioning management problem does not come from pyvisa-py, I found a solution to add a processing of these packets in the file "…/site-packages/pyvisa/usbtmc.c", in the read() function of the USBTMC class :
when receiving the usbtmc frame (resp = raw_read(…)), I read the data length field directly in the usbtmc header and I receive the missing packets (resp += raw_read(…)) as long as the usbtmc frame is not the expected length.

It takes just 3 lines of code (between "### DEB PATCH" and "### END PATCH"):
=========================================================================
try:
    resp = raw_read(recv_chunk + header_size + max_padding)

    ### DEB PATCH                                                             # processing if partitioned frame :
    usbtmc_size = header_size + int.from_bytes(resp[4:8], byteorder='little') # calculating the number of bytes expected
    while len(resp) < usbtmc_size:                                            # as long as the frame is incomplete...
        resp += raw_read(recv_chunk + header_size + max_padding)              # ...we get a packet
    ### END PATCH

    response = BulkInMessage.from_bytes(resp)
except (usb.core.USBError, ValueError):
    # Abort failed Bulk-IN operation.
    self._abort_bulk_in(self._btag)
    raise

received.extend(response.data)
=========================================================================

This partitioned frame handling in usbtmc.c works very well. I tested it with the Siglent SSA3032X and SVA1032X, as well as with a Keysight EDUX1002A (which does not partition its frames). I did this under Linux and under Windows, it works fine.

These 3 additional lines of code remain transparent if they are not needed but provide a great service in the event of partitioning poorly managed by the lower layers.

Although it is not the role of pyvisa-py to do this processing, do you think it is possible to integrate it into a future update of the pyvisa-py module?

Best regard

The modified read method of the USBTMC class:
[...]

Issue #422, opened 2024-03-28
[COM] Performance issue - slow transfer - USB


0xDEADBEEF-ASAP:
Hi! SDS1202XE , USB, pyvisa, pyvisa-py, arch Linux (os not virtualized), python running in venv.

750.1 KiB / 1 min transfer rate.

Can it be improved? What can I change or check to make it faster?

![IMG_221](https://github.com/pyvisa/pyvisa-py/assets/108463841/2adab43d-94a8-464c-a42e-fe76c3f0add9)
![IMG_20240328_133533](https://github.com/pyvisa/pyvisa-py/assets/108463841/2926c786-a9dd-40a0-ae12-e1f3acc6ad69)
![IMG_20240328_132032](https://github.com/pyvisa/pyvisa-py/assets/108463841/a9e6d405-caba-4c63-ab64-8363418f30f7)

MatthieuDartiailh:
Can you post the output of ptvisa-info and try to downgrade to 0.7.1 ?
 
0.7.2 introduce a fix that may be responsible for the slow transfer you observe.

0xDEADBEEF-ASAP:
Hi! pyvisa-py 0.7.1 it download screenshot super fast but first screenshot is incomplete 200kb , next screenshots are 60kb - full screenshot should be ~750kb).


Pyvisa-info:

Machine Details:
   Platform ID:    Linux-6.8.2-arch2-1-x86_64-with-glibc2.39
   Processor:      

Python:
   Implementation: CPython
   Executable:     /home/arch/test2/.venv/bin/python
   Version:        3.11.8
   Compiler:       GCC 13.2.1 20230801
   Architecture:   ('x86', 64)
   Build:          Feb 12 2024 14:50:05 (#main)
   Unicode:        UCS4

PyVISA Version: 1.14.1

Backends:
   ivi:
      Version: 1.14.1 (bundled with PyVISA)
      Binary library: Not found
   py:
      Version: 0.7.1
      USB INSTR: Available via PyUSB (1.2.1). Backend: libusb1
      USB RAW: Available via PyUSB (1.2.1). Backend: libusb1
      TCPIP INSTR: Available 
         Resource discovery:
         - VXI-11: partial (psutil not installed)
         - hislip: disabled (zeroconf not installed)
      TCPIP SOCKET: Available 
      ASRL INSTR:
         Please install PySerial (>=3.0) to use this resource type.
         No module named 'serial'
      VICP INSTR:
[...]

Issue #420, opened 2024-03-14
MaxPacketSize from PR #417 might impact performance


sessl3r:
PR #417 has fixed a issue with Rigol scopes and likely others by using the reported MaxPacketSize of the USB endpoints.
A similar [PR ](https://github.com/dpenkler/linux-usbtmc/pull/9) was merged into usbtmc linux driver.

In the discussion over there it came up, that some Agilent scopes report a smaller MaxPacketSize then they can handle in reality and therefore this change decreased performance. I can not validate this and think it is bad practice to use those bigger packet sizes.

I completely dislike the way this was now solved in the usbtmc driver (default is 4k, user can overwrite to use the reported value). However I agree that if users have performance issues a overwrite option could be introduced for such quirks.

This is not a current issue - just a reminder to not loose this information.

PR #470, opened 2024-11-22
Support partial usbtmc reads and request the `size` amount of bytes from the device

Jimmyvandenbergh:
another update for reading the data back from the device. this enhances the PR #465 

changes explained below:
- `req = BulkInMessage.build_array(self._btag, size, None)` we want to request the amount of bytes from the device which user specifies. therefore the header may specify more bytes then in a single transaction `wMaxPacketSize` can be transmitted. the device may response with less bytes. 
- above is also needed when you want to support IEEE 488.2 arbitrary block reads. this normally will read 2 bytes #<number of bytes> then your read the number of bytes from the device which holds the exact amount of bytes the data package may hold. this to claim memory on the host for large data packets. see below image: 

image

-  `resp = raw_read(self.usb_recv_ep.wMaxPacketSize)` you may request `wMaxPacketSize` but the device may send less bytes than requested

- fix the indentation of the ` received_message.extend(received_transfer[: response.transfer_size])` this should always remove the alignment bytes instead of only on `EOM`

- updated the code to request `x` amount of `wMaxPacketSize` to be read instead of `wMaxPacketSize` this to improve readout speeds. 

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,8 @@ PyVISA-py Changelog
   has been read (see specification), and only expects a header on the first packet received.
 - fix usbtmc implementation to properly discard the alignment bytes
   ensuring only the actual data (`transfer_size`) is retained in the message PR #465 
+- Implemented partial USBTMC message functionality that allows reading the amount of bytes 
+  specified by host.
 - add support for VI_ATTR_SUPPRESS_END_EN for USB resources PR #449
 
 0.7.2 (07/03/2024)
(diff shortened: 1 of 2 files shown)


Jimmyvandenbergh:
Speed performance for a command that send a large amount of data back: 
backend:
- NI-VISA: ~2.406 seconds 
- pyvisa-py small chunks: ~3.275 seconds  current implementation in 0.7.3 (and pre large chunk implementation)  
- pyvisa-py large chunks: ~2.413 seconds

Jimmyvandenbergh:
I think, at least, following issues are impacted and should benefit from this PR. 

- this should solve issues when reading data from instruments occur after updating to 0.7.2 or later 
- reading binary data when data may be larger than 1 message or does not complete in the first read() (transfer size is not met) of the device 


#472
#468
#463
#436
#422 
#420
#414
#318

Issue #474, opened 2024-12-06
unable to list_resources() on Mac: TypeError


mattjaffe42:
I have a Siglent SDS1104X-E that I am able to connect to and read data from using a more manual approach from [here](https://github.com/FPiorski/SDS1104X-E_socket_waveform/blob/master/main.py). However, I am unable to connect using `pyvisa` with the `pyvisa-py` backend:

```
import pyvisa
rm = pyvisa.ResourceManager('@py')
rm.list_resources()
```

gives the following error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[60], line 3
      1 import pyvisa
      2 rm = pyvisa.ResourceManager('@py')
----> 3 rm.list_resources()

File [~/opt/anaconda3/envs/jaffe-lab-0/lib/python3.13/site-packages/pyvisa/highlevel.py:3137](http://localhost:8888/lab/tree/OneDrive%20-%20Montana%20State%20University/Documents%20-%20Jaffe%20Lab/matt/dailys/24/12/~/opt/anaconda3/envs/jaffe-lab-0/lib/python3.13/site-packages/pyvisa/highlevel.py#line=3136), in ResourceManager.list_resources(self, query)
   3089 def list_resources(self, query: str = "?*::INSTR") -> Tuple[str, ...]:
   3090     r"""Return a tuple of all connected devices matching query.
   3091 
   3092     Notes
   (...)
   3134 
   3135     """
-> 3137     return self.visalib.list_resources(self.session, query)

File [~/opt/anaconda3/envs/jaffe-lab-0/lib/python3.13/site-packages/pyvisa_py/highlevel.py:487](http://localhost:8888/lab/tree/OneDrive%20-%20Montana%20State%20University/Documents%20-%20Jaffe%20Lab/matt/dailys/24/12/~/opt/anaconda3/envs/jaffe-lab-0/lib/python3.13/site-packages/pyvisa_py/highlevel.py#line=486), in PyVisaLibrary.list_resources(self, session, query)
    485 resources: List[str] = []
    486 for key, st in sessions.Session.iter_valid_session_classes():
--> 487     resources += st.list_resources()
    489 return rname.filter(resources, query)

[...]

PR #478, opened 2024-12-17
tcpip: ignore network with a netmask set to None

MatthieuDartiailh:
For such netmask, the mask is 32 bits long per psutil docs.


MatthieuDartiailh:
Closes #474 

Can you test @mattjaffe42 ?

mattjaffe42:
@MatthieuDartiailh tested, and confirmed resolves the error!

PR #484, opened 2025-01-22
Prologix support for both USB and ethernet versions

bobmcnamara:
Support added for Prologix GPIB adapters, both USB and ethernet.  Examples of usage:

`    import pyvisa`
`    rm = pyvisa.ResourceManager()`
`    prlgx = rm.open_resource("PRLGX-TCPIP::169.254.1.80::INTFC")`
`    instr = rm.open_resource("GPIB::17::INSTR")`
`    instr.query("*IDN?")`

Or for USB version:

`    import pyvisa`
`    rm = pyvisa.ResourceManager()`
`    prlgx = rm.open_resource("PRLGX-ASRL::/dev/cu.usbserialXXX::INTFC")`
`    instr = rm.open_resource("GPIB::17::INSTR")`
`    instr.query("*IDN?")`

- [n/a] Closes # (insert issue number if relevant)
- [X] Executed ``ruff check . && ruff format -c . --check`` with no errors
- [ToDo] The change is fully covered by automated unit tests
- [ToDo] Documented in docs/ as appropriate

### DIFF
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -9,7 +9,7 @@
 
 import ctypes  # Used for missing bindings not ideal
 from bisect import bisect
-from typing import Any, Iterator, List, Optional, Tuple, Type, Union
+from typing import Any, Iterator, List, Tuple, Type, Union
 
 from pyvisa import attributes, constants
 from pyvisa.constants import ResourceAttribute, StatusCode
@@ -36,7 +36,7 @@ class GPIBSessionDispatch(Session):
         resource_manager_session: VISARMSession,
         resource_name: str,
         parsed=None,
-        open_timeout: Optional[int] = None,
+        open_timeout: int | None = None,
     ) -> Session:
         newcls: Type
 
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -52,8 +52,7 @@ class GPIBSessionDispatch(Session):
 
 
 def make_unavailable(msg: str) -> Type:
-    """
-    This creates a fake session class that raises a ValueError if instantiated.
+    """Creates a fake session class that raises a ValueError if instantiated.
 
     We can't use Session.register_unavailable() because we need to be able to
     first check if a GPIB "board" has been registered in prologix.BOARDS.
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -22,7 +22,7 @@ from .sessions import Session, UnavailableSession, UnknownAttribute, VISARMSessi
 
 @Session.register(constants.InterfaceType.gpib, "INSTR")
 class GPIBSessionDispatch(Session):
-    """dispatch to the proper class based on entries in prologix.BOARDS.
+    """Dispatch to the proper class based on entries in prologix.BOARDS.
 
     Uses the __new__ method to intercept the creation of the instance of a
     GPIB session.  If parsed.board is found in prologix.BOARDS, create an
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -9,14 +9,71 @@
 
 import ctypes  # Used for missing bindings not ideal
 from bisect import bisect
-from typing import Any, Iterator, List, Tuple, Union
+from typing import Any, Iterator, List, Optional, Tuple, Type, Union
 
 from pyvisa import attributes, constants
 from pyvisa.constants import ResourceAttribute, StatusCode
-from pyvisa.rname import GPIBInstr, GPIBIntfc
+from pyvisa.rname import GPIBInstr, GPIBIntfc, parse_resource_name
 
+from . import prologix
 from .common import LOGGER
-from .sessions import Session, UnknownAttribute
+from .sessions import Session, UnavailableSession, UnknownAttribute, VISARMSession
+
+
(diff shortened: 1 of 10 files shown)


MatthieuDartiailh:
Thanks for the PR. I will do my best to review it by early next week.

pyvisa_py/prologix.py
```diff
@@ -0,0 +1,397 @@
...
+    """
+    This class is instantiated for PRLGX-ASRL<n>::INTFC resources.
+    """
+
+    # Override parsed to take into account the fact that this class is only
+    # used for specific kinds of resources
+    parsed: rname.PrlgxASRLIntfc  # type: ignore[assignment]
+
+    @staticmethod
+    def list_resources() -> List[str]:
```
MatthieuDartiailh, line 169:
```suggestion
    def list_resources() -> list[str]:
```

pyvisa_py/prologix.py
```diff
@@ -0,0 +1,397 @@
...
+
+    def after_parsing(self) -> None:
+        self.interface = pyserial.serial_for_url(
+            ("COM" if IS_WIN else "") + self.parsed.serial_device,
+            timeout=self.timeout,
+            write_timeout=self.timeout,
+            baudrate=115200,
+        )
+        # self.interface.write_termination = "\n"
+        # self.write_termination = "\n"
```
MatthieuDartiailh, line 183: To be removed ?

pyvisa_py/prologix.py
```diff
@@ -0,0 +1,397 @@
...
+    # want it to be registered in the _session_classes array, but we still
+    # need to define session_type to make the set_attribute machinery work.
+    session_type = (constants.InterfaceType.gpib, "INSTR")
+
+    # Override parsed to take into account the fact that this
+    # class is only used for a specific kind of resource
+    parsed: rname.GPIBInstr
+
+    @staticmethod
+    def list_resources() -> List[str]:
```
MatthieuDartiailh, line 239:
```suggestion
    def list_resources() -> list[str]:
```

pyvisa_py/prologix.py
```diff
@@ -0,0 +1,397 @@
...
+        self.interface = None
+        return StatusCode.success
+
+    def read(self, count: int) -> Tuple[bytes, StatusCode]:
+        if self.interface is None or self.interface.interface is None:
+            raise errors.InvalidSession()
+
+        self.interface.gpib_addr = self.gpib_addr
+
+        return self.interface.read(count)
```
MatthieuDartiailh, line 266: Looking at this code I think we need a lock on the interface.

MatthieuDartiailh (commented):
I left a bunch of minor stylistic comments.

However I think there is one more serious issue related to thread safety that need to be addressed. When setting the gpib_addr of the interface we need to be sure that no thread will modify it before the interface is done doing the job we need it to do.

pyvisa_py/prologix.py
bobmcnamara: Good catch, I've added a lock

pyvisa_py/prologix.py
bobmcnamara: Those lines have been removed.

pyvisa_py/prologix.py
```diff
@@ -0,0 +1,397 @@
...
+
+        """
+        if self.interface is None:
+            raise errors.InvalidSession()
+
+        try:
+            # use select to wait for write ready
+            rd, _wr, _ = select.select([self.interface], [self.interface], [])
+            if rd:
+                self.clear()
```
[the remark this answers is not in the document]
bobmcnamara: Yes, I've added comments about discarding stale read data.

MatthieuDartiailh:
The mypy failure is logical while waiting for https://github.com/pyvisa/pyvisa/pull/887 the test failure however needs to be addressed (simply by adding the new missing resources).

pyvisa_py/serial.py
MatthieuDartiailh, line 1: This should not be reverted.

pyvisa_py/gpib.py
```diff
@@ -9,20 +9,78 @@
...
+        parsed=None,
+        open_timeout: int | None = None,
+    ) -> Session:
+        newcls: Type
+
+        if parsed is None:
+            parsed = parse_resource_name(resource_name)
+
+        if parsed.board in prologix.BOARDS:
+            newcls = prologix.PrologixInstrSession
```
MatthieuDartiailh, line 47: Since the boards are registered through the opening of a resource via the resource manager, I think it would make sense to store them on the library linked to the resource manager. To enable this we could modify the signature of the base Session class which we can do since it is not user facing.

MatthieuDartiailh (commented):
First sorry for the slow review process.

Overall this looks good. I would just like to avoid storing the prologix interfaces globally. Can you make this change ?

pyvisa_py/gpib.py
bobmcnamara: Thanks for the suggestion, it's definitely better to eliminate the module-level global.  I moved it to a class var instead (in the prologix._PrologixIntfcSession class).

MatthieuDartiailh:
I think that rebasing should fix the typing issue. For the test, I will let you have a look. I will try to merge on Tuesday and make a release after.

PR #490, opened 2025-02-21
Fix large data transfer using USBTMC

### LINKED ISSUE #211: Having problems with large data transfers (NI vs pyvisa-py)
After looking at successful transfers from my RIgol ds1054z (screen captures 30kbytes in size) on the NI backend, and comparing them with my Ubuntu pyvisa-py backend... I believe pyusb is messing-up somewhere. I have been looking at wireshark captures, and the Windows 10 NI captures compare very well with my Ubuntu ones with the exception being the configurations of the bulk end points (I believe). I am able to see the USBTMC data packets and they all appear proper. That leaves the configuration of the BULK-in and BULK-out endpoints.
My packet lengths are 64 to 128 bytes on pyusb and 40 to 60 bytes on Windows 10. The data payloads are the same. When I look under /dev I can see that pyusb was able to disconnect the kernel driver, but maybe something was done to the interface which messes-up my large data transfers.
Any ideas would be appreciated.

2xB:
Transfer of large (e.g. binary) data did not work before due to an accidentally overwritten variable. This is solved in this commit by changing the variable name.

- [x] Closes #211

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ PyVISA-py Changelog
 - add support for VI_ATTR_SUPPRESS_END_EN for USB resources PR #449
 - support open_timeout for TCPIP hislip resources PR #430
 - fix serial flow control configuration PR #483
+- Fix large data transfer using USBTMC PR #490
 
 0.7.2 (07/03/2024)
 ------------------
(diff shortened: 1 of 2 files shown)


2xB:
This issue in our case is significant since https://github.com/pyvisa/pyvisa-py/pull/417 because in our case that reduced the transmission length below the length of our message. It did however also exist before, just occurred only for much larger data.

MatthieuDartiailh:
Thanks !!!

Issue #54, opened 2015-08-19
RIGOL DM3068 TCP/IP ERROR IN LINUX


zikako74:
Hello, 

I'm going crazy in order to communicate with the DMM pyvisa-py on TCPIP:

2015-08-20 00:09:39,056 - pyvisa - DEBUG - Created library wrapper for unset
2015-08-20 00:09:39,056 - pyvisa - DEBUG - Created ResourceManager with session 2052846
2015-08-20 00:09:39,056 - pyvisa - DEBUG - TCPIP0::192.168.1.144::INSTR - opening ...
2015-08-20 00:09:39,091 - pyvisa - DEBUG - TCPIP0::192.168.1.144::INSTR - is open with session 7697116
2015-08-20 00:09:39,106 - pyvisa - DEBUG - TCPIP0::192.168.1.144::INSTR - reading 20480 bytes (last status <StatusCode.success_max_count_read: 1073676294>)

2015-08-20 00:09:48,769 - pyvisa - DEBUG - TCPIP0::192.168.1.144::INSTR - closing
Exception NotImplementedError: NotImplementedError() in <bound method TCPIPInstrument.__del__ of <'TCPIPInstrument'(u'TCPIP0::192.168.1.144::inst0::INSTR')>> ignored

the python script is this:

import visa
visa.log_to_screen()
rm = visa.ResourceManager("@py")
inst = rm.open_resource("TCPIP0::192.168.1.144::INSTR")
print(inst.query("*IDN?"))

I'm used Linux 64bit and WindowsXp with python 2.7 32bit, but the error is the same.
I sniffed the network traffic and does not receive the correct response, the protocol VXI11 reports a timeout error.
With original software and NI-VISA library work fine, not with PyVisa and the same NI-VISA.

This is the debug with last pyvisa-py and python3

2015-08-20 00:23:17,237 - pyvisa - DEBUG - SerialSession was correctly imported.
2015-08-20 00:23:17,254 - pyvisa - DEBUG - USBSession and USBRawSession were correctly imported.
2015-08-20 00:23:17,264 - pyvisa - DEBUG - TCPIPSession was not imported cannot import name 'TCPIPSession'.
2015-08-20 00:23:17,266 - pyvisa - DEBUG - GPIBSession was not imported No module named 'gpib'.
2015-08-20 00:23:17,267 - pyvisa - DEBUG - Created library wrapper for unset
[...]

PR #496, opened 2025-04-04
Fix VXI-11 Default Lock Timeout (e.g. for Rigol DM3068)

sam-sw:
As explained at length in #54 and https://optics.eee.nottingham.ac.uk/vxi11/, pyvisa-py fails to communicate over Ethernet (via VXI-11) with the Rigol DM3068 digital multimeter. This is because the DM3068 expects a VXI-11 lock timeout of 0, not the default 10000ms used in pyvisa-py (which results in failure, and a 'Device Locked' response).

This commit alters where that 10000ms default is set, so it is now easily changeable by the end-user. For example:
```python
import pyvisa
rm = pyvisa.ResourceManager()
dm3068 = rm.open_resource('TCPIP::rigol-dm3068-hostname::INSTR')
# default lock_timeout is still 10000ms at this point
rm.visalib.sessions[dm3068.session].lock_timeout = 0
# can now communicate successfully with the DM3068
```

### DIFF
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -409,8 +409,8 @@ class TCPIPInstrVxi11(Session):
     #: Maximum size of a chunk of data in bytes.
     max_recv_size: int
 
-    #: Time to wait before erroring with a timeout when trying to acquire a lock
-    lock_timeout: int
+    #: Time to wait in ms before erroring with a timeout when trying to acquire a lock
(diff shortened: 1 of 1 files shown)


MatthieuDartiailh:
The change looks good but I think we need to add some documentation about this workaround to make it discoverable. The fact that VISA does not expose this VXI-11 parameter is tricky. An entry in the FAQ may be the best way to go.

PR #531, opened 2025-10-23
Add write_termination byte for TCPIP socket session

bollenn:
For some of our power supply devices it seems the write_termination is a mandatory field but it is not added for TCPIP protocol, so hereby I'm adding it.


bollenn:
@MatthieuDartiailh seems you are right, seems I overlooked a bug in our power supply definitions... thx for the tip

PR #567, opened 2026-03-17
Implement viTerminate() for HiSLIP sessions

### LINKED ISSUE #566: Implement viTerminate() for HiSLIP sessions (cross-thread I/O cancellation)
pyvisa-py has no implementation of `viTerminate()` — the `VisaLibraryBase.terminate()` method raises `NotImplementedError`. For HiSLIP sessions, this means there is no way to cancel a pending I/O operation from another thread.

 1. A SCPI command like `*WAI` or a binary data query can block for seconds to minutes while the instrument completes an operation.
 2. The thread blocked in `socket.recv_into()` inside `hislip.Instrument.receive()` cannot be interrupted.
 3. The only workaround is `socket.shutdown(SHUT_RDWR)` from another thread.

The VISA spec defines `viTerminate()` to handle exactly this case: cancel a pending synchronous operation so it returns `VI_ERROR_ABORT`, without destroying the session.

  ## Proposed solution

  Implement `viTerminate()` for `TCPIPInstrHiSLIP` using a cancellable socket wrapper.

  ### CancellableSocket

  Wrap the HiSLIP sync socket so that `recv_into()` uses `select()` to watch both the
  data socket and a cancel pipe (`socket.socketpair()`). When `cancel()` is called from
  another thread, the blocked `select()` returns immediately.

  ```python
  class CancellableSocket:
      def __init__(self, sock):
          self._sock = sock
          self._cancel_r, self._cancel_w = socket.socketpair()
          ...

      def recv_into(self, buffer, nbytes=0, flags=0):
          readable, _, _ = select.select(
              [self._sock, self._cancel_r], [], [], timeout
          )
          if self._cancel_r in readable:
              self._drain_cancel()
              raise HiSLIPInterruptedError(0)
          return self._sock.recv_into(buffer, nbytes, flags)

      def cancel(self):
          self._cancel_w.send(b"\x00")
```

 This is a drop-in replacement — all existing code that calls sock.recv_into() (including receive_exact_into, RxHeader, receive_flush) gets cancellation support transparently.

### PR DESCRIPTION
**Pull request by jjgilles99**
- [x] Closes #566
- [x] Executed ``ruff check && ruff format --check`` with no errors

## Summary

Implement `viTerminate()` for HiSLIP TCPIP sessions — the VISA function for
cross-thread I/O cancellation that was previously unimplemented
(`VisaLibraryBase.terminate()` raised `NotImplementedError`).

### Problem

When a thread is blocked in `hislip.Instrument.receive()` waiting for a long
SCPI operation, there is no way to cancel it from another thread. The only
workaround is `socket.shutdown(SHUT_RDWR)`, which destroys the TCP connection
and forces a full session reconnect.

### Approach

**CancellableSocket** — wraps the HiSLIP sync socket so that `recv_into()`
uses `select()` to watch both the data socket and a cancel pipe
(`socket.socketpair()`). When `cancel()` is called from another thread, the
blocked `select()` returns immediately with `HiSLIPInterruptedError` (mapped
to `StatusCode.error_abort`).

This is a drop-in replacement — all existing code that calls `sock.recv_into()`
(`receive_exact_into`, `RxHeader`, `receive_flush`) gets cancellation support
transparently.

**Two-phase terminate:**
- `terminate()` — instant, thread-safe. Writes to the cancel pipe.
- `complete_terminate()` — called after the read thread exits. Performs a full
  HiSLIP device clear per IVI-6.1 to re-sync the protocol.

**Interrupted message handling** — `_next_data_header()` now properly handles
`Interrupted` messages per IVI-6.1 Section 3.8 (previously silently discarded).

### Changes

| File | Change |
|------|--------|
| `protocols/hislip.py` | `CancellableSocket`, `HiSLIPInterruptedError`, `terminate()` + `complete_terminate()`, `Interrupted` handling |
| `tcpip.py` | `TCPIPInstrHiSLIP.terminate()` + `complete_terminate()`, catch abort in `read()` |
| `sessions.py` | `Session.terminate()` base method |
| `highlevel.py` | `PyVisaLibrary.terminate()` dispatcher |
| `testsuite/test_hislip_terminate.py` | 9 unit tests |

~270 lines added, 0 removed. All existing tests pass.

## Test plan

- [x] `TestCancellableSocket` — normal recv, cross-thread cancel, cancel-before-recv, drain+reuse, timeout, delegation, cancel priority
- [x] `TestHiSLIPInterruptedInHeader` — Interrupted message parsed correctly
- [x] `TestTerminateConcurrency` — cancel unblocks `receive_exact_into` from another thread
- [x] Full existing test suite passes (47 passed, 0 failed)
- [x] Live instrument test (would need a HiSLIP instrument)

### DIFF
--- a/pyvisa_py/highlevel.py
+++ b/pyvisa_py/highlevel.py
@@ -225,7 +225,7 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
         self,
         session: VISASession,
         degree: None,
-        job_id: VISAJobID,
+        job_id: VISAJobID | None,
     ) -> StatusCode:
         """Request a VISA session to terminate normal execution of an operation.
 
@@ -237,7 +237,7 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
             Unique logical identifier to a session.
         degree : None
             Not used in this version of the VISA specification.
-        job_id : VISAJobID
+        job_id : VISAJobID | None
             Specifies an operation identifier.  If None, aborts all calls
             on the specified session.
 
--- a/pyvisa_py/highlevel.py
+++ b/pyvisa_py/highlevel.py
@@ -23,7 +23,7 @@ from typing import (
 
 from pyvisa import constants, highlevel, rname
 from pyvisa.constants import StatusCode
-from pyvisa.typing import VISAEventContext, VISARMSession, VISASession
+from pyvisa.typing import VISAEventContext, VISAJobID, VISARMSession, VISASession
 from pyvisa.util import DebugInfo, LibraryPath
 
 from .common import LOGGER
@@ -225,15 +225,11 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
         self,
         session: VISASession,
         degree: None,
-        job_id,
+        job_id: VISAJobID,
     ) -> StatusCode:
         """Request a VISA session to terminate normal execution of an operation.
 
         Corresponds to viTerminate function of the VISA library.
-
-        For HiSLIP sessions, this cancels a pending read by signalling the
-        CancellableSocket's cancel pipe.  The blocked read returns with
-        StatusCode.error_abort.
 
         Parameters
         ----------
@@ -241,7 +237,7 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
             Unique logical identifier to a session.
         degree : None
             Not used in this version of the VISA specification.
-        job_id :
+        job_id : VISAJobID
             Specifies an operation identifier.  If None, aborts all calls
             on the specified session.
 
@@ -255,7 +251,7 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
             sess = self.sessions[session]
         except KeyError:
             return self.handle_return_value(session, StatusCode.error_invalid_object)
-        return self.handle_return_value(session, sess.terminate())
+        return self.handle_return_value(session, sess.terminate(job_id))
 
     def flush(
         self, session: VISASession, mask: constants.BufferOperation
--- a/pyvisa_py/highlevel.py
+++ b/pyvisa_py/highlevel.py
@@ -221,6 +221,42 @@ class PyVisaLibrary(highlevel.VisaLibraryBase):
             return self.handle_return_value(session, StatusCode.error_invalid_object)
         return self.handle_return_value(session, sess.clear())
 
+    def terminate(
+        self,
+        session: VISASession,
+        degree: None,
+        job_id,
+    ) -> StatusCode:
+        """Request a VISA session to terminate normal execution of an operation.
+
+        Corresponds to viTerminate function of the VISA library.
+
+        For HiSLIP sessions, this cancels a pending read by signalling the
+        CancellableSocket's cancel pipe.  The blocked read returns with
+        StatusCode.error_abort.
+
+        Parameters
+        ----------
+        session : VISASession
+            Unique logical identifier to a session.
+        degree : None
+            Not used in this version of the VISA specification.
+        job_id :
+            Specifies an operation identifier.  If None, aborts all calls
+            on the specified session.
+
+        Returns
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+        return self.handle_return_value(session, sess.terminate())
+
     def flush(
         self, session: VISASession, mask: constants.BufferOperation
     ) -> StatusCode:
--- a/pyvisa_py/protocols/hislip.py
+++ b/pyvisa_py/protocols/hislip.py
@@ -7,6 +7,7 @@ http://www.ivifoundation.org/downloads/Class%20Specifications/IVI-6.1_HiSLIP-1.1
 import select
 import socket
 import struct
+import threading
 import time
 from typing import Dict, Optional, Tuple
 
@@ -496,6 +497,7 @@ class Instrument:
         self._last_message_id: Optional[int] = None
         self._msg_type: str = ""
         self._payload_remaining: int = 0
+        self._receiving = threading.Event()
 
     # ================ #
     # MEMBER FUNCTIONS #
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: pyvisa_py/highlevel.py [hunk 1]
```diff
@@ -221,6 +221,42 @@ def clear(self, session: VISASession) -> StatusCode:
...
+        self,
+        session: VISASession,
+        degree: None,
+        job_id,
+    ) -> StatusCode:
+        """Request a VISA session to terminate normal execution of an operation.
+
+        Corresponds to viTerminate function of the VISA library.
+
+        For HiSLIP sessions, this cancels a pending read by signalling the
```
MatthieuDartiailh: This comment does not belong here.

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,42 @@ def clear(self, session: VISASession) -> StatusCode:
             return self.handle_return_value(session, StatusCode.error_invalid_object)
         return self.handle_return_value(session, sess.clear())
 
+    def terminate(
+        self,
+        session: VISASession,
+        degree: None,
```
MatthieuDartiailh, line 227: Missing type hint

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,42 @@ def clear(self, session: VISASession) -> StatusCode:
...
+        job_id,
```
MatthieuDartiailh: Missing type hint

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,42 @@ def clear(self, session: VISASession) -> StatusCode:
...
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+        return self.handle_return_value(session, sess.terminate())
```
MatthieuDartiailh: The job_id should be passed to the session even if currently only None is supported.

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,42 @@ def clear(self, session: VISASession) -> StatusCode:
             return self.handle_return_value(session, StatusCode.error_invalid_object)
         return self.handle_return_value(session, sess.clear())
 
+    def terminate(
```
MatthieuDartiailh, line 224: Reading the VISA specs (https://www.ni.com/docs/fr-FR/bundle/ni-visa-api-ref/page/ni-visa-api-ref/viterminate.html), viTerminate appears to apply mostly in the context of asynchronous operation.
Did you test your workflow with another implementation of VISA (NI, Keysight) to see if canceling a synchronous read does work ?

pyvisa_py/tcpip.py
```diff
@@ -331,6 +334,46 @@ def read_stb(self) -> Tuple[int, StatusCode]:
...
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        interface = cast(hislip.Instrument, self.interface)
+        interface.terminate()
+        return StatusCode.success
+
+    def complete_terminate(self) -> StatusCode:
```
MatthieuDartiailh: While I understand the need for this, since there  is no vi equivalent function I do not get how users are expected to call this.

Shouldn't it be called when read is interrupted by the exception signaling terminate was called?

pyvisa_py/protocols/hislip.py
```diff
@@ -116,6 +117,77 @@
...
+    thread, any blocked recv_into() returns immediately with
+    HiSLIPInterruptedError.
+
+    This implements the "self-pipe trick" for viTerminate() support.
+    """
+
+    def __init__(self, sock: socket.socket) -> None:
+        family, type_, proto = sock.family, sock.type, sock.proto
+        timeout = sock.gettimeout()
+        fd = sock.detach()
```
MatthieuDartiailh, line 150: Don't you need to copy the socket options over ? In any case a comment explaining how socket options are handled would be welcome.

MatthieuDartiailh (commented):
Looks nice but I have several questions.

pyvisa_py/tcpip.py
jjgilles99: thats a good suggestion, I'll move it to get called automatically when we catch the cancel

pyvisa_py/protocols/hislip.py
jjgilles99: socket options are persevered by the open file descriptor, I'll add a comment about it

pyvisa_py/highlevel.py
jjgilles99: Our use case is an asyncio-based instrument control system where synchronous VISA reads (viRead) run in background threads via asyncio.to_thread(). When the orchestrator needs to halt a long-running measurement (e.g., abort a slow VNA sweep), halt() is called from the event loop thread while a worker thread is blocked in viRead. Without viTerminate, the only option is `socket.shutdown(SHUT_RDWR)` which destroys the TCP session. The VISA spec supports this via `viTerminate(vi, VI_NULL, VI_NULL)` — "abort _any_ calls in the current process executing on the specified vi." we'd probably use `viReadAsync/viWriteAsync` if they were implemented but this is the minimal changeset to make our usecase cleaner

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,38 @@ def clear(self, session: VISASession) -> StatusCode:
...
+        self,
+        session: VISASession,
+        degree: None,
+        job_id: VISAJobID,
```
MatthieuDartiailh:
```suggestion
        job_id: VISAJobID | None,
```
Since you use mainly None

pyvisa_py/highlevel.py
MatthieuDartiailh: Could you test using Keysight VISA (lighter than NI) that mainstream implementation do behave in the way we expect ?

pyvisa_py/highlevel.py
```diff
@@ -221,6 +221,38 @@ def clear(self, session: VISASession) -> StatusCode:
...
+
+        Corresponds to viTerminate function of the VISA library.
+
+        Parameters
+        ----------
+        session : VISASession
+            Unique logical identifier to a session.
+        degree : None
+            Not used in this version of the VISA specification.
+        job_id : VISAJobID
```
MatthieuDartiailh:
```suggestion
        job_id : VISAJobID | None
```

pyvisa_py/sessions.py
```diff
@@ -419,6 +419,25 @@ def clear(self) -> StatusCode:
         """
         return StatusCode.error_nonsupported_operation
 
+    def terminate(self, job_id: VISAJobID = None) -> StatusCode:
```
MatthieuDartiailh:
```suggestion
    def terminate(self, job_id: VISAJobID | None = None) -> StatusCode:
```

pyvisa_py/sessions.py
```diff
@@ -419,6 +419,25 @@ def clear(self) -> StatusCode:
...
+        """Cancel a pending I/O operation.
+
+        Corresponds to viTerminate function of the VISA library.
+
+        Parameters
+        ----------
+        job_id : VISAJobID, optional
```
MatthieuDartiailh:
```suggestion
        job_id : VISAJobID | None, optional
```

pyvisa_py/tcpip.py
```diff
@@ -331,6 +338,32 @@ def read_stb(self) -> Tuple[int, StatusCode]:
 
         return stb, errorcode
 
+    def terminate(self, job_id: VISAJobID = None) -> StatusCode:
```
MatthieuDartiailh:
```suggestion
    def terminate(self, job_id: VISAJobID | None = None) -> StatusCode:
```

pyvisa_py/tcpip.py
```diff
@@ -331,6 +338,32 @@ def read_stb(self) -> Tuple[int, StatusCode]:
...
+        Corresponds to viTerminate function of the VISA library.
+
+        Thread-safe: may be called from any thread while another thread is
+        blocked in read().  The blocked read() will return with
+        StatusCode.error_abort and automatically reset the HiSLIP protocol
+        state so the session is ready for further I/O.
+
+        Parameters
+        ----------
+        job_id : VISAJobID, optional
```
MatthieuDartiailh:
```suggestion
        job_id : VISAJobID | None, optional
```

MatthieuDartiailh (commented):
Couple of comments regarding typing. And I would really like a confirmation of the expected behavior on mainstream implementations. I may be able to do it at some point but not before a month or so, so anybody is welcome to beat me to it.

jjgilles99:
Yes I can test with the Keysight VISA implementation - is there any specific way you'd like me to upload evidence? Just python logs is fine? I can also test this works with a Keysight VNA if that helps

pyvisa_py/highlevel.py
jjgilles99: yes I can do that

MatthieuDartiailh:
Yes logs would be fine. I would just like some evidence that we won't be deviating from mainstream implementation.

jjgilles99:
Looks like the Keysight VISA does not implement the terminate `None` condition as you expected - `viTerminate` just returns immediately and does not actually cancel the task in the other thread. This implementation does work as expected to cancel all outstanding tasks if we would like to go forward with the implementation. If you don't want to stray from the other implementations for some reason, I could either 1. implement the async read and write functions or 2. fork this repo for our internal use and not make the open source contributions
[logs_keysight_visa.txt](https://github.com/user-attachments/files/26223961/logs_keysight_visa.txt)
[logs_pyvisa_py.txt](https://github.com/user-attachments/files/26223964/logs_pyvisa_py.txt)
[test_viterminate_live.py](https://github.com/user-attachments/files/26223966/test_viterminate_live.py)

MatthieuDartiailh:
You did a good job implementing the feature and Indeed it can be interpreted as part of the specs so I am rather in favor of merging.

(Furthermore, the Async API from VISA does not match well to Python in my opinion and if asyncio based API is ever added to pyvisa I would rather rely on asyncio directly in pyvisa-py but that is a rather vague provision.)

However I would like a note ine the documentation calling out the fact that relying on that API may not be as portable as one may expect from VISA.

jjgilles99:
Thanks for the review! I added some notes about portability and added a fix for when calling terminate with no outstanding calls

MatthieuDartiailh (approved):

MatthieuDartiailh:
Sorry for the delay. Can you add an entry to the release notes ? (Sorry for not mentionning it earlier)

jjgilles99:
should be already in there? https://github.com/pyvisa/pyvisa-py/pull/567/changes#diff-aac07aefec754086f5c344e2c233862f6a611c739b5647fd456c1b77d3577c79R8 or did you mean somewhere else?

MatthieuDartiailh:
I meant in CHANGES

jjgilles99:
image
a different one than this ?

MatthieuDartiailh:
Sorry. I had onlyca quick look earlier and the file was folded... all good

GuidoKiener:
The code by AI "sounds" reasonable, but the handling of interrupted messages has nothing to do with asynchronous commands like viWriteAsync, viReadAsync, and viTerminate. Instead, the (Async)Interrupted messages are only sent by the server to clean up stale messages (see appendix A.2 Fast Client and chapter 3.1.1). The interrupted messages are required for the so-called "Message Exchange Protocol".
BTW a second thread is needed on client side to handle the race between "AsyncInterrupted" and "Interrupted" messages.

MatthieuDartiailh:
@GuidoKiener I will be the first to admit I am not a specialist of the hislip specification. This PR addressed the user issue and the code seemed fine so I merged. 
You are welcome to contribute a patch to improve the current situation. Note that currently true async operation are not used by PyVISA.

PR #547, opened 2025-12-13
Event handlers

SimonMerrett:
- [x] Closes #544 
- [ ] Executed ``black . && isort -c . && flake8`` with no errors
- [ ] The change is fully covered by automated unit tests
- [ ] Documented in docs/ as appropriate
- [ ] Added an entry to the CHANGES file

As requested by @MatthieuDartiailh in #544 this PR attempts to implement the suggestions in the gist linked in that issue, as a basis for introducing TCPIP::INST event handlers. I do not expect to close #544 but link to it anyway as this is the issue of interest for this PR.

I'm not sure if there is a branch other than main I should be targeting with this PR. 

I have not added an entry to CHANGES as I fully expect this to require a lot of edits before it works and passes testing.

I have not written unit tests but have included a test script based on the Gemini suggestion. I have not run the formatting tools as I'm not sure how to and pylance in VSCode showed no issues with typing nor formatting. I realise the scope of these tools is not likely a complete match with pylance.

PR #577, opened 2026-04-28
Add event handling for TCPIP instruments, allowing SRQ interrupts. Closes #544

### LINKED ISSUE #544: Event handlers - is it possible for TCPIP INSTR resource?
I previously used NI backend but never want to let another NI software run on my new, clean machine. So I tried to use event handlers with @py backend but found that these are apparently not yet implemented in pyvisa-py.  Is that correct?

If so, I'm willing to help implement with some supervision. Is this something the maintainers can give?

SimonMerrett:
My second attempt at providing the event handling I would like to see for instruments capable of sending SRQ service requests.  @MatthieuDartiailh helpfully advised on the approach and reviewed my previous PR but I did not and do not have enough Python skill to take that review into action. I have done my best in this PR to provide the feature but actually was able to test it on hardware which took 5 measurements at 1Hz and returned SRQ. I have attached Wireshark pcap showing the communication of a successful test (measurement results are not retrieved during the test).
[pyvisa-py_srq.zip](https://github.com/user-attachments/files/27182581/pyvisa-py_srq.zip)

I used an agentic workflow in opencode to plan, code, test, review. Some of the wider-reaching changes were bound up in reviewer recommendations for robustness, such as compatibility "tweaks" in attributes.py. Everything is up for editing but with this PR I am happy that it should pass some degree of testing (the agents were testing against the built-in tests and added their own for the event aspects). I am also more confident because I was able to take my script for my instrument and test the functionality was there, that measurements were taken, SRQ was sent properly as far as I can tell (not a hidden python thread polling the instrument status registers, like one of the earlier agents tried to sneak in).

Please go easy on the feedback and feel free to edit! If you agree that this is closer to the mark than https://github.com/pyvisa/pyvisa-py/pull/547 I will close that PR. 

- [X] Closes #544 (maybe)
- [X] The change is fully covered by automated unit tests (I belive so)
- [X] Documented in docs/ as appropriate ( index.rst has been updated)


pyvisa_py/attributes.py
```diff
@@ -10,6 +10,9 @@
 from pyvisa import constants
 from pyvisa.attributes import AttrVI_ATTR_TCPIP_KEEPALIVE as former_keepalive
 
+# vicp may not exist in older pyvisa versions
+_vicp = getattr(constants.InterfaceType, "vicp", None)
+
```
MatthieuDartiailh: We enforce a recent enough version of pyvisa (or we should I will have to double check). So this is not needed.

pyvisa_py/attributes.py
```diff
@@ -27,5 +30,5 @@ class AttrVI_ATTR_TCPIP_KEEPALIVE(former_keepalive):
     resources = [
         (constants.InterfaceType.tcpip, "SOCKET"),
         (constants.InterfaceType.tcpip, "INSTR"),
-        (constants.InterfaceType.vicp, "INSTR"),
+        *([(_vicp, "INSTR")] if _vicp is not None else []),
```
MatthieuDartiailh: Revert

MatthieuDartiailh:
Thanks for the PR given the size of the changes ot will take me time to do a thorough review. Do not hesitate to ping me if I go silent for too long.

xyphro:
BTW: I really like that SRQs support gets added. Was missing it!

pyvisa_py/events.py
```diff
@@ -0,0 +1,286 @@
...
+import time
+from dataclasses import dataclass, field
+from typing import Any, Callable
+
+from pyvisa import constants
+
+from .common import LOGGER
+
+
+@dataclass
```
MatthieuDartiailh: This should be marked as frozen since the intent is for instances to be immutable. Additionally we could have slot=True

pyvisa_py/events.py
```diff
@@ -0,0 +1,286 @@
...
+        with self._cond:
+            if event_type is None:
+                self._deque.clear()
+            else:
+                kept = [ctx for ctx in self._deque if ctx.event_type != event_type]
+                self._deque.clear()
+                self._deque.extend(kept)
+
+
+HandlerCallback = Callable[[Any, constants.EventType, int, Any], None]
```
MatthieuDartiailh: A docstring explaining the meaning of each argument would be welcomed.

pyvisa_py/events.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class HandlerRegistry:
+    """Thread-safe registry of user-installed event handlers."""
+
+    def __init__(self) -> None:
+        self._lock = threading.RLock()
+        # event_type -> list of (handler, user_handle)
+        self._handlers: dict[
```
MatthieuDartiailh: Using a defaultdict here would simplify the code

pyvisa_py/events.py
```diff
@@ -0,0 +1,286 @@
...
+
+        """
+        with self._lock:
+            handlers = list(self._handlers.get(event_type, []))
+
+        for handler, user_handle in handlers:
+            try:
+                handler(session, event_type, context_id, user_handle)
+            except Exception:
+                LOGGER.exception("Exception in event handler for %s", event_type)
```
MatthieuDartiailh: I wonder if we should not call all handlers and raise an exception group with all raised exceptions rather than just logging.
Another option would be to issue a warning which would be more visible than a log entry.

pyvisa_py/events.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class EventState:
+    """Per-session container for event enablement, queuing, and handlers."""
+
+    def __init__(self) -> None:
+        # {event_type: {mechanism_int}}
+        self._lock = threading.RLock()
+        self.enabled: dict[constants.EventType, set[int]] = {}
```
MatthieuDartiailh: Using a Flag here rather than unnamed integer values in a set would be much more readable.

pyvisa_py/highlevel.py
```diff
@@ -7,6 +7,8 @@
 
 """
 
+from __future__ import annotations
```
MatthieuDartiailh: If this does not bring anything I would rather not add it.

pyvisa_py/highlevel.py
```diff
@@ -790,6 +794,47 @@ def unlock(self, session: VISASession) -> StatusCode:
...
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+        if event_type not in sess._supported_event_types:
```
MatthieuDartiailh, line 829: Some blank lines between logic block would improve the readability.

pyvisa_py/sessions.py
```diff
@@ -366,6 +374,43 @@ def after_parsing(self) -> None:
...
+        This method is called by transport-specific monitor threads when an
+        SRQ (or other asynchronous signal) is detected.
+        """
+        queue_enabled, handler_enabled = self._event_state.get_delivery_mechanisms(
+            event_type
+        )
+        if queue_enabled:
+            self._event_state.queue.put(ctx)
+        if handler_enabled:
+            session_handle = getattr(self, "_session_handle", self)
```
MatthieuDartiailh: This looks a bit suspicious. What is the rational.

pyvisa_py/sessions.py
```diff
@@ -366,6 +374,43 @@ def after_parsing(self) -> None:
...
+            self._event_state.registry.fire(event_type, session_handle, ctx.context_id)
+
+    def _start_srq_monitor(self) -> StatusCode:
```
MatthieuDartiailh: The name seems a bit narrow since other types of event can be monitored

MatthieuDartiailh (commented):
This is a first pass. I think one of the main thing that could be improved is the tracking of the type of events for which we handle events. Using an int flag internally could simplify many things I think.

pyvisa_py/attributes.py
SimonMerrett: This should be removed now.

pyvisa_py/attributes.py
SimonMerrett: Reverted

pyvisa_py/events.py
SimonMerrett: Done. Both frozen and slots.

pyvisa_py/events.py
SimonMerrett: Added docstring. Please confirm if formatting is acceptable or I need to look at this.

pyvisa_py/events.py
SimonMerrett: defaultdict implemented

pyvisa_py/events.py
SimonMerrett: flags implemented. Following pyvisa flag values.

pyvisa_py/highlevel.py
SimonMerrett: Removed.

pyvisa_py/highlevel.py
SimonMerrett: Blank lines added.

pyvisa_py/sessions.py
SimonMerrett: I do not know but I asked for a review and this was the output 
```
1. Design intent: why the getattr fallback exists
_session_handle is assigned only once in the entire codebase — in highlevel.py:201 immediately after an instrument session is created:
visa_session = self._register(sess)
sess._session_handle = visa_session
The getattr(self, "_session_handle", self) on sessions.py:387 is defensive coding: the author knew _session_handle is not part of the Session base class __init__ and added a fallback in case it was missing. However, the fallback self (the Python Session object) is never the correct value to pass to a VISA event handler.
2. The Resource Manager session
The RM session is created in highlevel.py:524 by registering the PyVisaLibrary object itself. _session_handle is never set on it.
Relevance: None. The RM session lacks _event_state, _supported_event_types, and any event monitor. If event methods were called on it they would AttributeError long before reaching _fire_event. _fire_event can only be invoked on instrument sessions, and those always receive _session_handle.
3. The test fixture
lib_and_session creates a MagicMock, manually attaches _event_state and _supported_event_types, registers it with the library, but forgets to set _session_handle. The existing test test_vxi11_fire_event_handler_mechanism passes only because its handler ignores the first argument, masking the fact that the mock object (not the integer handle) is being passed.
Verdict: This is a test bug, not a code bug. The fallback accommodates an incomplete test fixture.
4. VISA spec compliance
The handler contract (per the skill context and upstream pyvisa conventions) is:
handler(session: VISASession, event_type: EventType, event_ctx: int, user_handle: Any) -> None
VISASession is an integer handle. Passing self (a Python object) violates this contract.
---
Recommendation
✅ Recommended: Option A
Replace getattr(self, "_session_handle", self) with a direct attribute access:
session_handle = self._session_handle
Rationale:
- Every session that can legitimately fire events is an instrument session created through highlevel.open(), which unconditionally sets _session_handle.
- The fallback self is spec-violating and serves only to hide the incomplete test fixture.
- Direct attribute access makes the invariant obviously correct: "a session firing events must have a VISA handle".
- Option B documents incorrect behavior; Option C introduces unnecessary indirection for a problem that is really a missing test fixture line.
---
Required Test Changes
1. Fix lib_and_session fixture (test_events.py:~345)  
   Add one line so the mock session mirrors what real highlevel.open() does:
      @pytest.fixture
   def lib_and_session():
       lib = PyVisaLibrary()
       sess = MagicMock()
       sess._event_state = EventState()
       sess._supported_event_types = {constants.EventType.service_request}
       sess._start_event_monitor.return_value = StatusCode.success
       session_id = lib._register(sess)
       sess._session_handle = session_id          # <-- ADD THIS
       return lib, sess, session_id
   
2. Strengthen test_vxi11_fire_event_handler_mechanism (test_events.py:~695)  
   Update the handler to capture the session argument and assert it is the integer handle (sid), not the mock object:
      def test_vxi11_fire_event_handler_mechanism(self, lib_and_session):
       lib, sess, sid = lib_and_session
       calls = []
       def my_handler(session, event_type, context_id, user_handle):
           calls.append((session, event_type, context_id, user_handle))
       # ... install and enable as before ...
       Session._fire_event(sess, constants.EventType.service_request, ctx)
       assert calls == [(sid, constants.EventType.service_request, 5555, "uh")]
   
No other tests require modification because they either use the queue mechanism (which does not touch _session_handle) or test EventState/highlevel directly without calling Session._fire_event.
```
So I implemented the recommendation. I will need more guidance if this is leading in the wrong direction.

pyvisa_py/sessions.py
SimonMerrett: Generalised the names to `_start_event_monitor` / `_stop_event_monitor`

SimonMerrett:
@MatthieuDartiailh  thanks for your review comments. I think I have addressed everything, although I cannot be confident I haven't introduced other mistakes. All pytest tests are passing but I haven't confirmed function on my test instrument yet. Please continue your review and I will confirm functionality asap.

SimonMerrett:
[pyvisa-py_srq-test_script_log.txt](https://github.com/user-attachments/files/27492426/pyvisa-py_srq-test_script_log.txt)
[test_pcap_after_commit-bdc446a.zip](https://github.com/user-attachments/files/27492428/test_pcap_after_commit-bdc446a.zip)
Just tested with my test instrument and had successful script output. pcap attached to assist with verification of the protocol.

SimonMerrett:
@MatthieuDartiailh just a polite nudge 🙂

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+                self._deque.clear()
+                self._deque.extend(kept)
+
+
+# HandlerCallback: callable invoked when a VISA event fires.
+#   Arg 0 (Any):                  session handle (vi)
+#   Arg 1 (constants.EventType):  the event type that fired
+#   Arg 2 (int):                  event context id
+#   Arg 3 (Any):                  user-supplied handle passed at install_handler time
+HandlerCallback = Callable[[Any, constants.EventType, int, Any], None]
```
MatthieuDartiailh: I feel like the session handle could be more strongly typed. It cannot an arbitrary object.
```suggestion
HandlerCallback = Callable[[Any, constants.EventType, int, Any], None]
"""Callable invoked when a VISA event fires.

Parameters
‐---------
Arg 0 : Any
    session handle (vi)
Arg 1 : constants.EventType
    The event type that fired
Arg 2 : int
    Event context id
Arg 3 : Any
    User-supplied handle passed at install_handler time.

"""
```

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+    SUSPEND == 7``), but it is accepted from ``constants.EventMechanism.all``
+    and is canonicalised to the three real flags on store so that bitwise
+    ``~`` works correctly.
+    """
+
+    NONE = 0
+    QUEUE = 1  # VI_QUEUE   (1)
+    HANDLER = 2  # VI_HNDLR   (2)
+    SUSPEND = 4  # VI_SUSPEND_HNDLR (4)
+    ALL = 0xFFFF  # VI_ALL_MECH (0xFFFF)
```
MatthieuDartiailh: Having a from_int method managing the conversion of 0xFFFF to the right flag superposition would be nicer to avoid the possibility of creating meaningless instances.

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+                    del self.enabled[event_type]
+                else:
+                    self.enabled[event_type] = new
+
+    def is_queue_enabled(self, event_type: constants.EventType) -> bool:
+        """Return whether queue delivery is enabled for *event_type*."""
+        with self._lock:
+            return (
+                self.enabled.get(event_type, EventMechanism.NONE) & EventMechanism.QUEUE
+            ) is not EventMechanism.NONE
```
MatthieuDartiailh:
```suggestion
            return  bool(
                self.enabled.get(event_type, EventMechanism.NONE) & EventMechanism.QUEUE
            ) 
```

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def is_handler_enabled(self, event_type: constants.EventType) -> bool:
+        """Return whether handler (callback) delivery is enabled for *event_type*."""
+        with self._lock:
+            return (
+                self.enabled.get(event_type, EventMechanism.NONE)
+                & EventMechanism.HANDLER
+            ) is not EventMechanism.NONE
```
MatthieuDartiailh:
```suggestion
            return bool(
                self.enabled.get(event_type, EventMechanism.NONE)
                & EventMechanism.HANDLER
            )
```

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+        """Return (queue_enabled, handler_enabled) for *event_type*.
+
+        The check is performed atomically under the state lock.
+        """
+        with self._lock:
+            mech = self.enabled.get(event_type, EventMechanism.NONE)
+            return (
+                (mech & EventMechanism.QUEUE) is not EventMechanism.NONE,
+                (mech & EventMechanism.HANDLER) is not EventMechanism.NONE,
+            )
```
MatthieuDartiailh, line 307:
```suggestion
            return (
                bool(mech & EventMechanism.QUEUE),
                bool(mech & EventMechanism.HANDLER),
            )
```

pyvisa_py/events.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def any_enabled(self) -> bool:
+        """Return ``True`` if any event type has any mechanism enabled."""
+        with self._lock:
+            return any(m is not EventMechanism.NONE for m in self.enabled.values())
+
+    def should_monitor(self) -> bool:
+        """Convenience alias for :meth:`any_enabled`."""
+        return self.any_enabled()
```
MatthieuDartiailh: This looks unused and could be removed.

pyvisa_py/highlevel.py
```diff
@@ -790,6 +792,48 @@ def unlock(self, session: VISASession) -> StatusCode:
...
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+
+        if event_type not in sess._supported_event_types:
+            return self.handle_return_value(session, StatusCode.error_invalid_event)
+        sess._event_state.enable(event_type, mechanism)
+        status = sess._start_event_monitor()
+        if status != StatusCode.success:
+            sess._event_state.disable(event_type, mechanism)
+            return self.handle_return_value(session, status)
+        return self.handle_return_value(session, StatusCode.success)
```
MatthieuDartiailh, line 838:
```suggestion
        if event_type not in sess._supported_event_types:
            return self.handle_return_value(session, StatusCode.error_invalid_event)

        sess._event_state.enable(event_type, mechanism)
        status = sess._start_event_monitor()
        if status != StatusCode.success:
            sess._event_state.disable(event_type, mechanism)
            return self.handle_return_value(session, status)

        return self.handle_return_value(session, StatusCode.success)
```

pyvisa_py/highlevel.py
```diff
@@ -815,7 +859,19 @@ def disable_event(
...
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+
+        if event_type == constants.EventType.all_enabled:
+            for et in list(sess._event_state.enabled.keys()):
+                sess._event_state.disable(et, mechanism)
+        else:
+            sess._event_state.disable(event_type, mechanism)
+        if not sess._event_state.any_enabled():
+            sess._stop_event_monitor()
+        return self.handle_return_value(session, StatusCode.success)
```
MatthieuDartiailh, line 879:
```suggestion
        if event_type == constants.EventType.all_enabled:
            for et in list(sess._event_state.enabled.keys()):
                sess._event_state.disable(et, mechanism)
        else:
            sess._event_state.disable(event_type, mechanism)

        if not sess._event_state.any_enabled():
            sess._stop_event_monitor()

        return self.handle_return_value(session, StatusCode.success)
```

pyvisa_py/highlevel.py
```diff
@@ -842,4 +898,147 @@ def discard_events(
...
-        return StatusCode.error_nonimplemented_operation
+        try:
+            sess = self.sessions[session]
+        except KeyError:
+            return self.handle_return_value(session, StatusCode.error_invalid_object)
+
+        mech = int(mechanism)
+        if mech == int(constants.EventMechanism.all) or mech & int(
+            constants.EventMechanism.queue
+        ):
```
MatthieuDartiailh: This should use the flag.

pyvisa_py/highlevel.py
```diff
@@ -842,4 +898,147 @@ def discard_events(
...
+            et = None if event_type == constants.EventType.all_enabled else event_type
+            sess._event_state.queue.discard_all(et)
+        return self.handle_return_value(session, StatusCode.success)
```
MatthieuDartiailh, line 916:
```suggestion
            sess._event_state.queue.discard_all(et)

        return self.handle_return_value(session, StatusCode.success)
```

SimonMerrett:
@MatthieuDartiailh am I getting closer with this?

SimonMerrett:
Sorry, didn't mean to close.

pyvisa_py/protocols/vxi11.py
```diff
@@ -354,11 +369,207 @@ def create_intr_chan(self, host_addr, host_port, prog_num, prog_vers, prog_famil
...
             self.unpacker.unpack_device_error,
         )
 
     def destroy_intr_chan(self):
         return self.make_call(
             DESTROY_INTR_CHAN, None, None, self.unpacker.unpack_device_error
         )
+
+
+class SrqInterruptServer(rpc.UDPServer):
```
MatthieuDartiailh: What is the rationale for the UDP implementation? It is unused currently.

MatthieuDartiailh (commented):
This is looking nice. I still need a deeper look at the tests but I did not want to delay my feedback further. Sorry for the slow progress

SimonMerrett:
Thanks for the review work on this. I plan to address the comments this weekend.

SimonMerrett:
@MatthieuDartiailh I think I have addressed your comments.

pyvisa_py/protocols/vxi11.py
SimonMerrett: This has been removed.

pyvisa_py/highlevel.py
SimonMerrett: Now uses flag, if I understand correctly.

pyvisa_py/events.py
SimonMerrett: Have added a from_int class method

pyvisa_py/events.py
SimonMerrett: now uses a VISASession type

CHANGES
```diff
@@ -15,6 +15,14 @@ PyVISA-py Changelog
   A second addressed issue is that timeout values never decrement to 0. A timeout value 
   of 0 is undefined in VXI-11 standard. It can mean "timeout immediately if no data 
   is in buffer" or "block permanently until transfer is finished".
+- Implement the VISA event subsystem for VXI-11 (TCPIP::INSTR) resources:
+  `viEnableEvent`, `viDisableEvent`, `viDiscardEvents`, `viWaitOnEvent`,
+  `viInstallHandler`, and `viUninstallHandler`. SRQ (service request) events
+  are now supported via both queue-based (`wait_on_event`) and handler-based
+  (`install_handler`) delivery. A daemon thread runs an ONC RPC UDP interrupt
```
```suggestion
  (`install_handler`) delivery. A daemon thread runs an ONC RPC TCP interrupt
```

CHANGES
```diff
@@ -15,6 +15,14 @@ PyVISA-py Changelog
...
+  server to receive VXI-11 `DEVICE_INTR_SRQ` callbacks. This also fixes the
+  `create_intr_chan` XDR packer in `protocols/vxi11.py`.
+  Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now.
```
MatthieuDartiailh: Please reference the PR number

MatthieuDartiailh (commented):
LGTM

minor comments and typing issues to be fixed then I will merge

SimonMerrett:
Thanks for your patience @MatthieuDartiailh . I am happy to make those changes. I noticed that the mypy checks failed with comments:
```
pyvisa_py/highlevel.py:947: error: Incompatible return value type (got "tuple[EventType, int, StatusCode]", expected "tuple[EventType, VISAEventContext, StatusCode]")  [return-value]
pyvisa_py/highlevel.py:956: error: Incompatible return value type (got "tuple[EventType, int, StatusCode]", expected "tuple[EventType, VISAEventContext, StatusCode]")  [return-value]
pyvisa_py/highlevel.py:961: error: Incompatible return value type (got "tuple[EventType, int, StatusCode]", expected "tuple[EventType, VISAEventContext, StatusCode]")  [return-value]
```
Would you like me to change `highlevel.py` as follows:

1. Lines 947-951 (invalid session case):
_Change from:_
`return (in_event_type, 0, self.handle_return_value(session, StatusCode.error_invalid_object))`
_To:_
`return (in_event_type, VISAEventContext(0), self.handle_return_value(session, StatusCode.error_invalid_object))`

2. Lines 956-960 (timeout case):
_Change from:_
`return (in_event_type, 0, self.handle_return_value(session, StatusCode.error_timeout))`
_To:_
`return (in_event_type, VISAEventContext(0), self.handle_return_value(session, StatusCode.error_timeout))`

3. Line 961 (success case):
_Change from:_
`return ctx.event_type, ctx.context_id, StatusCode.success`
_To:_
`return ctx.event_type, VISAEventContext(ctx.context_id), StatusCode.success`

MatthieuDartiailh:
The fixes you propose look fine to me.

SimonMerrett:
@MatthieuDartiailh ready to merge

MatthieuDartiailh (approved):

PR #587, opened 2026-06-01
Add support for NI GPIB-ENET/100 bridge.

bytewarrior:
## Summary

Adds a pure-Python session driver for the **NI GPIB-ENET/100** Ethernet-to-GPIB bridge, modeled after the existing Prologix support. Once the companion pyvisa changes are merged, one can

```python
rm.open_resource("NI-ENET100-TCPIP0::192.0.2.42::INTFC")
inst = rm.open_resource("GPIB0::16::INSTR")  # dispatched through the bridge
print(inst.query("*IDN?"))
```

with no additional dependencies (stdlib `socket` / `struct` / `threading` only).

## Depends on

[pyvisa#980 – Add InterfaceType.ni_enet100_tcpip, NIEnet100TCPIPIntfc rname class and Resource subclass](https://github.com/pyvisa/pyvisa/pull/980).

Until that PR is merged, `from pyvisa_py import nienet100` raises `ImportError`, which `highlevel.py` swallows at debug level (mirrors the `vicp` pattern). All NIENET100 tests skip cleanly in that case, so this PR remains importable today; it just stays inert.

## What's added

- **Wire protocol** (`pyvisa_py/protocols/nienet100.py`)
- **Discovery** (`pyvisa_py/protocols/nienet100_discovery.py`) — UDP broadcast on the local broadcast domain.
- **Session layer** (`pyvisa_py/nienet100.py`) — `NIEnet100TCPIPIntfcSession` for `NI-ENET100-TCPIP::INTFC`, `NIEnet100InstrSession` for `GPIB::INSTR` resources that route through a registered bridge.
- **Dispatch hook** — chains in front of `gpib.py`'s `GPIBSessionDispatch` and falls back to it for non-NIENET100 boards. Coexists with Prologix: load order is `prologix → gpib → nienet100`, dispatch order at `open_resource` is `NIENET100 → Prologix → linux-gpib`. Users wire each bridge to a distinct board number.
- **Module wiring** in `pyvisa_py/highlevel.py` (one new `try/except` block).
- **Tests** — offline unit tests for the wire layer, discovery, and SRQ verbs; assisted (hardware-gated) tests for the wire layer and full pyvisa-stack session layer.

## Testing

**Offline** (always runs, no hardware needed):
- `pyvisa_py/testsuite/test_nienet100.py` (~25 tests) — wire framing, status header parsing, open/close sequences, SRQ verbs.
- `pyvisa_py/testsuite/test_nienet100_discovery.py` — discovery frame parsers + UDP loop with mocked sockets.

**Hardware-gated** (`pyvisa_py/testsuite/nienet100_assisted_tests/`) — opt-in via env vars:
```
PYVISA_TEST_NIENET100_HOST  # IP or hostname of the bridge
PYVISA_TEST_GPIB_PAD        # primary address of an instrument on the bus
PYVISA_TEST_GPIB_SAD        # (optional) secondary address
PYVISA_TEST_IDN_VENDOR      # (optional) substring expected in *IDN?
PYVISA_TEST_GPIB_TERM       # (optional) write/read termination — defaults to \n
```
Without `PYVISA_TEST_NIENET100_HOST` set, every assisted test skips cleanly. Verified locally against a real GPIB-ENET/100.

## Out of scope / known limitations

These are documented in the code and listed here so reviewers don't have to chase them:

- **Six of the seven `gpib_control_ren` modes** return `error_nonsupported_operation` — only `VI_GPIB_REN_DEASSERT_GTL` (which is by far the common case) is wired. The other six need an `ibsre` verb whose wire frame is not yet reverse-engineered. Prologix exposes none of these modes, so this is already a strict improvement.
- **No SRQ event delivery to pyvisa's event model** — `ibwait` is implemented at the wire layer but the bridge to pyvisa's `enable_event`/`wait_on_event` machinery is not.
- **`ibrsp` with `cnt != 1`** raises `NIEnet100ProtocolError`. The wire spec lets the bridge cram multi-byte STB blocks into a single response; in practice every device we've seen sends a single STB byte. Will relax when an instrument shows otherwise.
- **Multi-chunk `ibrd` data fragmentation** — the parser handles it (accumulates into `payload` across data chunks until END or final-status), but has not been validated against very large device responses.
- **`*IDN?` instrument tests assume `\n` termination** because pyvisa's `\r\n` default chokes several older GPIB instruments. The `PYVISA_TEST_GPIB_TERM` env var overrides.

The bridge itself has two intrinsic quirks worth noting for users:
- **Built-in timeout minimum ~3 s.** Per-call `tmo_ms` in `ibrd` does not lower this floor — short pyvisa timeouts (e.g. `inst.timeout = 200`) surface as the configured `error_timeout` but after ~3 s wall-clock, not 200 ms. The socket-level timeout ceiling is sized to accommodate.
- **Limited concurrent sessions.** Rapid open/close cycles can put the bridge into a state where it stops accepting new TCP connections for ~60 s. Practical workaround: serialize bridge sessions, or wait/power-cycle.

## Reviewer guidance

The commit history is structured for sequential review — each commit is self-contained and the rough phases are:

1. Wire-protocol primitives (chunks, status headers, command framing).
2. Open / close sequences and Device-I/O verbs.
3. Session-layer integration into pyvisa-py (`pyvisa_py/nienet100.py`, dispatch hook, module loader).
4. SRQ verbs and lazy wait/control socket lifecycle.
5. Discovery (frame primitives, UDP loop, `list_resources` wiring).
6. Offline tests.
7. Hardware-gated tests.
8. Wire-level fixes uncovered by hardware testing (`ibrsp` combined chunk, `ibrd` no-END path, IbcTMO rejection, status-header chunk wrapping).
9. Session-layer fixes uncovered by full-stack hardware testing (resource-string mapping, board-key type, timeout handling, test robustness).

The protocol was reverse-engineered from wire captures and the `gpibhlpr.dll` decompilate; key wire-layer details (status-header chunking, `ibrsp` combined-chunk format, `ibrd` two-chunk model, `tmo_ms` semantics) are documented inline at their use sites.

- [X] Executed ruff with no errors
- [X] The change is fully covered by automated unit tests

still open, will continue after initial feedback:

### DIFF
--- a/pyvisa_py/gpib_dispatch.py
+++ b/pyvisa_py/gpib_dispatch.py
@@ -99,6 +99,7 @@ class GPIBInstrDispatch(Session):
 # Priorities: lower values are consulted first. A backend bound to specific
 # boards (Prologix, NI-ENET/100 bridge) must out-rank the native driver,
 # which is the catch-all fallback and therefore comes last.
+_PRIORITY_BRIDGE = 10
 _PRIORITY_PROLOGIX = 20
 _PRIORITY_NATIVE = 100
 
@@ -153,6 +154,22 @@ def register_builtin_backends() -> None:
     global _builtins_registered
     if _builtins_registered:
         return
+
+    # NI GPIB-ENET/100 bridge: claims boards bound to a bridge INTFC. Highest
+    # precedence so a board registered to a bridge is never shadowed.
+    try:
+        from . import nienet100
+    except Exception as e:
+        LOGGER.debug("NI-ENET/100 GPIB::INSTR backend not registered: %s", e)
+    else:
+        register_backend(
+            _board_resolver(
+                nienet100._NIEnet100IntfcSession.boards,
+                nienet100.NIEnet100InstrSession,
+            ),
+            priority=_PRIORITY_BRIDGE,
+            label="ni-enet100",
+        )
 
     # Prologix controller: claims boards bound to a Prologix interface.
     try:
--- a/pyvisa_py/gpib_dispatch.py
+++ b/pyvisa_py/gpib_dispatch.py
@@ -28,7 +28,8 @@ from pyvisa import rname
 from pyvisa.constants import StatusCode
 from pyvisa.typing import VISARMSession
 
-from .sessions import OpenError, Session
+from .common import LOGGER
+from .sessions import OpenError, Session, UnavailableSession
 
 #: A resolver inspects a parsed resource name and returns the session class
 #: that should handle it, or ``None`` if the resource is not served by this
@@ -92,3 +93,91 @@ class GPIBInstrDispatch(Session):
                 )
 
         raise OpenError(StatusCode.error_resource_not_found)
+
+
+# --- Built-in backend wiring ------------------------------------------------
+# Priorities: lower values are consulted first. A backend bound to specific
+# boards (Prologix, NI-ENET/100 bridge) must out-rank the native driver,
+# which is the catch-all fallback and therefore comes last.
+_PRIORITY_PROLOGIX = 20
+_PRIORITY_NATIVE = 100
+
+#: Guards :func:`register_builtin_backends` so repeated calls (e.g. one per
+#: ResourceManager) do not append duplicate resolvers.
+_builtins_registered = False
+
+
+def _board_resolver(boards: dict, session_cls: Type[Session]) -> GPIBInstrResolver:
+    """Build a resolver that claims a resource only for registered boards.
+
+    The ``boards`` mapping is captured by reference and queried at dispatch
+    time, so boards registered after wiring (the normal case — INTFC
+    sessions populate it on open) are still seen.
+
+    """
+
+    def resolve(parsed: rname.ResourceName) -> Optional[Type[Session]]:
+        return session_cls if parsed.board in boards else None
+
+    return resolve
+
+
+def _make_native_unavailable(exc: Exception) -> Type[Session]:
+    """Return an unavailable session explaining the missing GPIB library.
+
+    ``gpib.py`` raises on import when neither linux-gpib nor gpib-ctypes is
+    present. Routing unmatched boards here preserves the previous
+    behaviour: a clear, actionable error at open time instead of a generic
+    "resource not found".
+
+    """
+
+    class _NativeGPIBUnavailable(UnavailableSession):
+        session_issue = (
+            "Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) "
+            "to use native GPIB::INSTR resources.\n%s" % exc
+        )
+
+    return _NativeGPIBUnavailable
+
+
+def register_builtin_backends() -> None:
+    """Wire pyvisa-py's in-tree GPIB::INSTR backends into the dispatcher.
+
+    Each backend is imported defensively: a backend whose import fails is
+    simply skipped (native gpib is additionally represented by an
+    unavailable session so its install hint survives). The call is
+    idempotent.
+
+    """
+    global _builtins_registered
+    if _builtins_registered:
+        return
+
+    # Prologix controller: claims boards bound to a Prologix interface.
+    try:
+        from . import prologix
+    except Exception as e:  # pragma: no cover - prologix import is robust
+        LOGGER.debug("Prologix GPIB::INSTR backend not registered: %s", e)
+    else:
+        register_backend(
+            _board_resolver(
+                prologix._PrologixIntfcSession.boards, prologix.PrologixInstrSession
+            ),
+            priority=_PRIORITY_PROLOGIX,
+            label="prologix",
+        )
+
+    # Native linux-gpib / gpib-ctypes: catch-all for any board not claimed
+    # above.
+    try:
+        from .gpib import GPIBSession
+    except Exception as e:
+        native_cls: Type[Session] = _make_native_unavailable(e)
+    else:
+        native_cls = GPIBSession
+    register_backend(
+        lambda parsed: native_cls, priority=_PRIORITY_NATIVE, label="gpib"
+    )
+
+    _builtins_registered = True
--- a/pyvisa_py/gpib_dispatch.py
+++ b/pyvisa_py/gpib_dispatch.py
@@ -19,7 +19,8 @@ library, so it can own the slot even when ``gpib.py`` fails to import.
 
 """
 
-from typing import Callable, List, Optional, Tuple, Type, cast
+from collections.abc import Callable
+from typing import cast
 
 from pyvisa import constants, rname
 from pyvisa.constants import StatusCode
@@ -34,14 +35,14 @@ from .sessions import OpenError, Session, UnavailableSession
 #: backend (e.g. the board number is not registered to it). Resources reach
 #: this dispatcher only via the ``(gpib, "INSTR")`` slot, so the parsed name
 #: is always a :class:`~pyvisa.rname.GPIBInstr`.
-GPIBInstrResolver = Callable[[GPIBInstr], Optional[Type[Session]]]
+GPIBInstrResolver = Callable[[GPIBInstr], type[Session] | None]
 
 #: Registered backends as ``(priority, label, resolver)``. Lower priority
 #: values are consulted first; :func:`register_backend` keeps the list
 #: sorted. The sort is stable, so insertion order breaks ties — import
 #: order therefore only affects equal-priority backends, never the
 #: correctness of the fallback chain.
-_GPIB_INSTR_BACKENDS: List[Tuple[int, str, GPIBInstrResolver]] = []
+_GPIB_INSTR_BACKENDS: list[tuple[int, str, GPIBInstrResolver]] = []
 
 
 def register_backend(
@@ -80,8 +81,8 @@ class GPIBInstrDispatch(Session):
         cls,
         resource_manager_session: VISARMSession,
         resource_name: str,
-        parsed: Optional[rname.ResourceName] = None,
-        open_timeout: Optional[int] = None,
+        parsed: rname.ResourceName | None = None,
+        open_timeout: int | None = None,
     ) -> Session:
         if parsed is None:
             parsed = rname.parse_resource_name(resource_name)
@@ -98,7 +99,7 @@ class GPIBInstrDispatch(Session):
         raise OpenError(StatusCode.error_resource_not_found)
 
     @staticmethod
-    def list_resources() -> List[str]:
+    def list_resources() -> list[str]:
         """List native GPIB::INSTR resources found on local boards.
 
         Only the native linux-gpib / gpib-ctypes listeners are enumerated:
@@ -132,7 +133,7 @@ _PRIORITY_NATIVE = 100
 _builtins_registered = False
 
 
-def _board_resolver(boards: dict, session_cls: Type[Session]) -> GPIBInstrResolver:
+def _board_resolver(boards: dict, session_cls: type[Session]) -> GPIBInstrResolver:
     """Build a resolver that claims a resource only for registered boards.
 
     The ``boards`` mapping is captured by reference and queried at dispatch
@@ -141,13 +142,13 @@ def _board_resolver(boards: dict, session_cls: Type[Session]) -> GPIBInstrResolv
 
     """
 
-    def resolve(parsed: GPIBInstr) -> Optional[Type[Session]]:
+    def resolve(parsed: GPIBInstr) -> type[Session] | None:
         return session_cls if parsed.board in boards else None
 
     return resolve
 
 
-def _make_native_unavailable(exc: Exception) -> Type[Session]:
+def _make_native_unavailable(exc: Exception) -> type[Session]:
     """Return an unavailable session explaining the missing GPIB library.
 
     ``gpib.py`` raises on import when neither linux-gpib nor gpib-ctypes is
@@ -214,7 +215,7 @@ def register_builtin_backends() -> None:
     try:
         from .gpib import GPIBSession
     except Exception as e:
-        native_cls: Type[Session] = _make_native_unavailable(e)
+        native_cls: type[Session] = _make_native_unavailable(e)
     else:
         native_cls = GPIBSession
     register_backend(lambda parsed: native_cls, priority=_PRIORITY_NATIVE, label="gpib")
--- a/pyvisa_py/gpib_dispatch.py
+++ b/pyvisa_py/gpib_dispatch.py
@@ -6,16 +6,13 @@ mutually exclusive backends — native linux-gpib / gpib-ctypes, a Prologix
 controller, or an NI GPIB-ENET/100 bridge. The session registry holds a
 single class per ``(InterfaceType.gpib, "INSTR")`` slot.
 
-This module owns the slot once and resolves the concrete session class at
-open time by consulting an ordered list of backend resolvers. Backends
-register themselves via :func:`register_backend` when they import
-successfully; a backend that fails to import is simply absent from the
-list, so the remaining backends keep working regardless of import order.
-
-The dispatcher is intentionally **not** registered in this module yet —
-that happens once the existing backends have been migrated onto
-:func:`register_backend`, so adding this module changes no behaviour on
-its own.
+This module owns the ``(gpib, "INSTR")`` slot once and resolves the
+concrete session class at open time by consulting an ordered list of
+backend resolvers. :func:`register_builtin_backends` imports each in-tree
+backend defensively and registers a resolver for it; a backend whose
+import fails is simply absent, so the remaining backends keep working
+regardless of import order. The module has no hard dependency on any GPIB
+library, so it can own the slot even when ``gpib.py`` fails to import.
 
 :copyright: 2026 by PyVISA-py Authors, see AUTHORS for more details.
 :license: MIT, see LICENSE for more details.
@@ -24,7 +21,7 @@ its own.
 
 from typing import Callable, List, Optional, Tuple, Type
 
-from pyvisa import rname
+from pyvisa import constants, rname
 from pyvisa.constants import StatusCode
 from pyvisa.typing import VISARMSession
 
@@ -66,6 +63,7 @@ def register_backend(
     _GPIB_INSTR_BACKENDS.sort(key=lambda item: item[0])
 
 
+@Session.register(constants.InterfaceType.gpib, "INSTR")
 class GPIBInstrDispatch(Session):
     """Resolve a ``GPIB::INSTR`` resource to the right backend session.
 
@@ -93,6 +91,27 @@ class GPIBInstrDispatch(Session):
                 )
 
         raise OpenError(StatusCode.error_resource_not_found)
+
+    @staticmethod
+    def list_resources() -> List[str]:
+        """List native GPIB::INSTR resources found on local boards.
+
+        Only the native linux-gpib / gpib-ctypes listeners are enumerated:
+        Prologix has no listing support, and NI GPIB-ENET/100 instruments
+        are discovered at the INTFC level. Returns an empty list when no
+        native GPIB library is installed.
+
+        """
+        try:
+            from .gpib import _find_listeners
+        except Exception:
+            return []
+        return [
+            "GPIB%d::%d::INSTR" % (board, pad)
+            if sad == 0
+            else "GPIB%d::%d::%d::INSTR" % (board, pad, sad - 0x60)
+            for board, pad, sad in _find_listeners()
+        ]
 
 
 # --- Built-in backend wiring ------------------------------------------------
--- a/pyvisa_py/gpib_dispatch.py
+++ b/pyvisa_py/gpib_dispatch.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+"""Central dispatch for GPIB INSTR resources across pyvisa-py backends.
+
+pyvisa-py can serve ``GPIB<n>::...::INSTR`` resources through several
+mutually exclusive backends — native linux-gpib / gpib-ctypes, a Prologix
+controller, or an NI GPIB-ENET/100 bridge. The session registry holds a
+single class per ``(InterfaceType.gpib, "INSTR")`` slot.
+
+This module owns the slot once and resolves the concrete session class at
+open time by consulting an ordered list of backend resolvers. Backends
+register themselves via :func:`register_backend` when they import
+successfully; a backend that fails to import is simply absent from the
+list, so the remaining backends keep working regardless of import order.
+
+The dispatcher is intentionally **not** registered in this module yet —
+that happens once the existing backends have been migrated onto
+:func:`register_backend`, so adding this module changes no behaviour on
+its own.
+
+:copyright: 2026 by PyVISA-py Authors, see AUTHORS for more details.
+:license: MIT, see LICENSE for more details.
+
+"""
+
+from typing import Callable, List, Optional, Tuple, Type
+
+from pyvisa import rname
+from pyvisa.constants import StatusCode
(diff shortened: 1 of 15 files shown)


MatthieuDartiailh:
Thanks for this. It will take me time to go through all of it. Do not hesitate to ping me if you do hear back within a couple of weeks.

bytewarrior:
Thanks, Matthieu. Your efforts are very much appreciated.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,472 @@
...
+        "InterfaceType.ni_enet100_tcpip; please update pyvisa."
+    ) from e
+
+try:
+    _RNAME_NIENET100_TCPIP_INTFC = rname.NIEnet100TCPIPIntfc
+except AttributeError as e:
+    raise ImportError(
+        "pyvisa-py NI GPIB-ENET/100 support requires pyvisa with "
+        "rname.NIEnet100TCPIPIntfc; please update pyvisa."
+    ) from e
```
MatthieuDartiailh: There is no reason to have 2 separate blocks here.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,472 @@
...
+
+    Holds the class-level ``boards`` registry that the GPIB dispatch hook
+    in :mod:`pyvisa_py.gpib` consults to route ``GPIB<n>::*::INSTR``
+    resources through the appropriate bridge.
+
+    The INTFC owns its own :class:`~pyvisa_py.protocols.nienet100.EnetConnection`
+    for the session lifetime. The connection acts as a connectivity sentinel
+    (the box rejects Device-I/O on stale sessions, so an open socket is a
+    reliable health signal). INSTR sessions do **not** share this connection;
+    they each open their own.
```
MatthieuDartiailh, line 44: Missing blank line at end of docstring

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,472 @@
...
+            data = self.interface.ibrd(tmo_ms=tmo_ms)
+        except nienet100.NIEnet100IOError as e:
+            return b"", _map_iberr_to_status(e.err)
+
+        # The wire-level ibrd always reads the full message (until EOI/EOS);
+        # truncate here if the caller's max-count is smaller. Extra bytes
+        # are dropped — ibrd has no resume semantics, so a caller that
+        # supplied too small a count loses the tail.
+        if len(data) > count:
+            return bytes(data[:count]), StatusCode.success_max_count_read
```
MatthieuDartiailh: This is not acceptable. The session should maintain an internal buffer to allow reading responses one byte at a time (useful in debugging).

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,472 @@
...
+
+        term_char, _ = self.get_attribute(ResourceAttribute.termchar)
+        term_en, _ = self.get_attribute(ResourceAttribute.termchar_enabled)
+        if term_en and term_char is not None and data and data[-1] == term_char:
+            return bytes(data), StatusCode.success_termination_character_read
+        return bytes(data), StatusCode.success
+
+    def clear(self) -> StatusCode:
```
MatthieuDartiailh, line 377: The internal buffer should also be cleared.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,472 @@
...
+    if iberr == nienet100.ERR_ECIC:
+        return StatusCode.error_not_cic
+    if iberr == nienet100.ERR_EARG:
+        return StatusCode.error_invalid_mode
+    if iberr == nienet100.ERR_ESAC:
+        return StatusCode.error_nonsupported_operation
+    return StatusCode.error_system_error
+
+
+# --- (gpib, INSTR) dispatch hook --------------------------------------------
```
MatthieuDartiailh: The following feels quite brittle since it relies heavily on the order of the imports. It may be better to have a central location registering GPIB::INSTR if any of the possible interfaces is functional and managing the dispatching in a single place.

MatthieuDartiailh (changes requested):

MatthieuDartiailh:
First quick round of review for high level stuff and interfaces.

pyvisa_py/nienet100.py
bytewarrior: I would suggest to reformat all of the docstrings then, in order to keep it consistent across the codebase.
I did it this way, as large portions of the code do not have blank lines either.

So the convention is to have a blank line at the end of each docstring?

pyvisa_py/nienet100.py
bytewarrior: Done, removed the second one and reworked the text message.

pyvisa_py/nienet100.py
bytewarrior: Understood. I have added an internal read buffer to the INSTR session. Reading responses byte-wise is now possible.

pyvisa_py/nienet100.py
bytewarrior: Done, as this followed directly from adding the read buffer.

pyvisa_py/nienet100.py
bytewarrior: Yes, this is exactly the point where I was unsure how to solve it.

This single place you're referring to is probably supposed to be in pyvisa, not pyvisa-py?

Thinking further, and as an alternative, one could perhaps add a dispatcher here in pyvisa-py, something like gpib_dispatch.py. This one would populate the (gpib, INSTR) slot once and for all without the cumbersome pop/re-register stuff.

It could also create an ordered candidate list of backends, to which each backend would contribute if imported successfully.

What do you think?

pyvisa_py/nienet100.py
MatthieuDartiailh: I follow numpy suggested format which does mandate a blank line at the end of a docstring.

pyvisa_py/nienet100.py
MatthieuDartiailh: My idea was indeed to have that single point of registration living in pyvisa-py.

pyvisa_py/nienet100.py
bytewarrior: Ok, I've added a blank line to all multi-line docstrings.

pyvisa_py/nienet100.py
bytewarrior: Ok, and I suppose it's okay to implement this inside this PR and not create a separate one?

MatthieuDartiailh: Yes

bytewarrior: Ok, so I went ahead and gave it a try. Instead of the earlier approach of popping and re-wrapping the registry at import time, the backends are now registered as priority-ordered resolvers:

- NI-ENET/100 bridge (priority 10): board bound to a NI GPIB/ENET-100 INTFC
- Prologix (priority 20): board bound to a Prologix controller
- native linux-gpib / gpib-ctypes (priority 100): catch-all fallback

`register_builtin_backends()` imports each backend defensively, so a backend whose import fails (e.g. gpib.py raising when no GPIB library is installed) is simply absent from the list. As a consequence, the remaining backends keep working. The module itself has no hard dependency on any GPIB library, so it can own the slot even when gpib.py fails to import.

Dispatch no longer depends on the order in which backends are imported, and adding another GPIB backend in the future is a one-line `register_backend()`.

I've yet come across another question: `list_resources()` still enumerates only native GPIB listeners (it delegates to `gpib._find_listeners()` exactly as the old `GPIBSessionDispatch.list_resources` did). Prologix has never been listed (the old `# FIXME` is still accurate), and GPIB-ENET/100s are discovered at the INTFC level via `NI-ENET100-TCPIP…::INTFC`. For now, I kept that behaviour identical rather than change discovery semantics.
Is this okay?

MatthieuDartiailh (commented):
A general comment, I did not go through everything yet. Since the supported Python version is 3.10 for new/edited code I would like to avoid the typing imports List, Tuple, etc

bytewarrior:
I've made two new commits-- one addressing the Python 3.10 compatibility/ syntax issue.
One more on the way as I noticed a few offline tests would not run on Linux.

pyvisa_py/gpib_dispatch.py
```diff
@@ -0,0 +1,223 @@
...
+from pyvisa.typing import VISARMSession
+
+from .common import LOGGER
+from .sessions import OpenError, Session, UnavailableSession
+
+#: A resolver inspects a parsed resource name and returns the session class
+#: that should handle it, or ``None`` if the resource is not served by this
+#: backend (e.g. the board number is not registered to it). Resources reach
+#: this dispatcher only via the ``(gpib, "INSTR")`` slot, so the parsed name
+#: is always a :class:`~pyvisa.rname.GPIBInstr`.
```
MatthieuDartiailh, line 37: I have been convinced that even though this format is supported by sphinx using standard docstring after the object is better in particular because it is better supported by IDE.

So at least in new contributions it would make sense to do it that way.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,452 @@
...
+# falls back when pyvicp is not installed). Users opening NI-ENET100-TCPIP
+# resources then see a clean "No class registered" error instead of a
+# cryptic AttributeError during session creation.
+try:
+    _IFACE_NIENET100_TCPIP = constants.InterfaceType.ni_enet100_tcpip
+except AttributeError as e:
+    raise ImportError(
+        "pyvisa-py NI GPIB-ENET/100 support requires pyvisa with "
+        "some definitions specific to nienet100; please update pyvisa."
+    ) from e
```
MatthieuDartiailh: In this case this should be simply handled by requesting a recent enough version of pyvisa in pyvisa-py.
For the development period it may be worth keeping but we should cleanup before the release.

pyvisa_py/protocols/nienet100.py
```diff
@@ -0,0 +1,1156 @@
+# -*- coding: utf-8 -*-
+"""Python implementation of the NI GPIB-ENET/100 wire protocol.
+
+This module talks the proprietary TCP protocol of the National Instruments
+GPIB-ENET/100 Ethernet-to-GPIB bridge. It is **not** compatible with the
+older GPIB-ENET (10 MBit/s, libnienet target), which uses a similar frame
+layout but different verb opcodes and a single-step open.
+
+Wire reference: ``work/GPIB-ENET-100_Protocol.md``.
```
MatthieuDartiailh: This reference does not exist.

pyvisa_py/testsuite/nienet100_assisted_tests/test_session.py
```diff
@@ -0,0 +1,192 @@
...
+    assert "0" in boards, "INTFC did not register board 0: boards=%r" % (
+        list(boards.keys()),
+    )
+
+
+# --- instrument tests (need PYVISA_TEST_GPIB_PAD) -------------------------
+
+
+@require_instrument
+def test_idn_query_via_pyvisa(inst):
```
MatthieuDartiailh, line 149: Could you extend this test to test reading the response using multiple read bytes to stress the use of the intermediate buffer.

MatthieuDartiailh (commented):
Some comments but we are close.

Thanks a mot for tour work.

pyvisa_py/nienet100.py
bytewarrior: Ok, I'll add a TODO here, and leave that until the end, when it is known which minimum version is required.

pyvisa_py/protocols/nienet100.py
bytewarrior: This is an overlooked reference to my protocol reverse engineering docs I did not catch. I'll reformulate and update the comment and remove the stale reference.

pyvisa_py/testsuite/nienet100_assisted_tests/test_session.py
bytewarrior: Done. Also, I've noticed the code would not clear the buffer in case of a scenario like:
- write and start to read the response
- leave tail bytes in the buffer
- write again, expect to read response from the second write

Hence, I've modified the write to clear the intermediate buffer as well, and added a test to capture this case.

bytewarrior:
I have found a problem with ibwait while running the tests many times in succession with real hardware. I probably have to change a few pieces regarding the ibwait/SRQ machinery. Working on it.

pyvisa_py/gpib_dispatch.py
bytewarrior: I'll format it as you wish, but I'm a bit uncertain now what to do. Should I change it to improve the IDE experience, or leave it as it is?

bytewarrior:
Ok, this was a bit more than I had expected. I've run into some issues while running the tests multiple times. Turns out my conclusions drawn from the disassembly alone were not entirely correct regarding ibwait/SRQ, in fact this led to removing significant portions of code dealing with 5003/5005 ports (as of now unused, never opened).

I decided to give the code this kind of cleanup. I hope you don't mind. I have addressed your comments as well of course.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,483 @@
...
+    def after_parsing(self) -> None:
+        # pyvisa open_timeout is in milliseconds; convert to seconds for the
+        # socket layer. ``None`` or 0 (VI_TMO_IMMEDIATE, the pyvisa default)
+        # means "use a sane default": a literal ~1 ms socket timeout is far
+        # too short for the multi-frame board open, so fall back to 10 s (as
+        # VXI-11 does) rather than ``max(0, 0.001)``.
+        if not self.open_timeout:
+            connect_timeout_s = 10.0
+        else:
+            connect_timeout_s = max(self.open_timeout / 1000.0, 0.001)
```
MatthieuDartiailh, line 161: This is not exactly wrong but open_timeout is specified as a mean to control the timeout on acquiring a lock when open the resource. It confuses all users. Here we deviate from the standard but match expectations. 
We can keep it I just wanted to point out this VISA quirk.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,483 @@
...
+
+        host = self.parsed.host_address
+        try:
+            self.interface = nienet100.EnetConnection(
+                host,
+                open_timeout=connect_timeout_s,
+                timeout=connect_timeout_s,
```
MatthieuDartiailh: This however is wrong I think, you should pass the value of the timeout attribute if set or a default value.

pyvisa_py/nienet100.py
```diff
@@ -0,0 +1,483 @@
...
+            return 0, StatusCode.error_connection_lost
+        # A new write starts a fresh exchange: drop any buffered-but-unread
+        # bytes left over from a partial read of a previous response so the
+        # stale tail cannot be prepended to the reply for this command.
+        self._read_buffer.clear()
+        try:
+            written = self.interface.ibwrt(data)
+        except nienet100.NIEnet100IOError as e:
+            return 0, _map_iberr_to_status(e.err)
+        return written, StatusCode.success
```
MatthieuDartiailh: I never checked what happen with other VISA implementation. Not reading the response usually get a query interrupted error from the device but here the issue is different. I am not sure it is our responsibility to deal with poorly managed message recovery in this way.

pyvisa_py/protocols/nienet100.py
```diff
@@ -0,0 +1,1070 @@
...
+    100e-3,
+    300e-3,
+    1.0,
+    3.0,
+    10.0,
+    30.0,
+    100.0,
+    300.0,
+    1000.0,
+)
```
MatthieuDartiailh: PR #601 is standardizing those so that we stop having bare integer all over the place. Could you use those once the PR has been merged ?

pyvisa_py/protocols/nienet100.py
```diff
@@ -0,0 +1,1070 @@
...
+        for diagnosing wire-protocol surprises against real hardware. Set
+        ``--log-cli-level=DEBUG`` on pytest to see the dumps, or attach a
+        handler to ``pyvisa_py.protocols.nienet100`` in your own code.
+
+        """
+        if self.main is None:
+            raise NIEnet100Error("main socket is not open")
+        data = self._recv_exactly(self.main, n)
+        if LOGGER.isEnabledFor(logging.DEBUG):
+            LOGGER.debug("← main: %s", data.hex())
```
MatthieuDartiailh: The Python logging module will take care of filtering logging for disabled levels, no need to do it manually.

MatthieuDartiailh (commented):
Sorry but this PR is huge. I did a review on the bulk of the code. I will need more time to go through the added tests.

pyvisa_py/protocols/nienet100.py
bytewarrior: Yes, of course.

MatthieuDartiailh: PR #601 has been merged so you can use the constants it defines.

bytewarrior: Done.

pyvisa_py/protocols/nienet100.py
bytewarrior: Basically yes. However, without the `if`, `data.hex()` will be called in a rather hot code path (recv/send), allocating memory for each conversion, even when debug logging is not enabled.

This is why the `if` was added.

Should I remove the two `if` statements anyway? It's only used in send/recv this way.

MatthieuDartiailh: You can directly request the formatting in the string using %x and this way you avoid the need to manually call hex.

bytewarrior: I'm probably doing something wrong here. I cannot directly use "%x", because `data` is a `bytes`-like object. "%x" on the other hand expects a _single integer_. What I wanted to do here is to have a "lazy" hex conversion.

Any more thoughts/ suggestions?

Should I just drop the `if` altogether and accept the performance penalty at this point? After all, I'm wondering if it makes a large difference anyway.

pyvisa_py/testsuite/test_nienet100_session.py
```diff
@@ -0,0 +1,70 @@
...
+# The session module raises ImportError on load when the upstream pyvisa
+# additions it depends on (InterfaceType.ni_enet100_tcpip) are missing; skip
+# the whole module cleanly in that case, mirroring the assisted suite.
+try:
+    from pyvisa_py import gpib_constants, nienet100 as ni
+    from pyvisa_py.protocols import nienet100 as proto
+except ImportError as _import_err:  # pragma: no cover - depends on pyvisa version
+    pytestmark = pytest.mark.skip(
+        reason="pyvisa-py NI GPIB-ENET/100 session layer unavailable: %s" % _import_err
+    )
```
MatthieuDartiailh: Same as in other places this will have to be cleaned up.

pyvisa_py/testsuite/test_nienet100.py
```diff
@@ -0,0 +1,629 @@
...
+from pyvisa_py import gpib_constants
+from pyvisa_py.protocols import nienet100
+
+# --- frame pack / unpack ----------------------------------------------------
+
+
+def test_pack_command_zeroes_unset_fields():
+    frame = nienet100.pack_command(0x04)
+    assert frame == b"\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+    assert len(frame) == nienet100.COMMAND_FRAME_SIZE
```
MatthieuDartiailh:
```suggestion
    assert len(frame) == nienet100.COMMAND_FRAME_SIZE
    assert frame == b"\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```

MatthieuDartiailh:
```suggestion
    assert len(frame) == nienet100.COMMAND_FRAME_SIZE
    assert frame == b"\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```

MatthieuDartiailh:
```suggestion
    assert len(frame) == nienet100.COMMAND_FRAME_SIZE
    assert frame == b"\x04\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
```

pyvisa_py/testsuite/nienet100_assisted_tests/test_session.py
```diff
@@ -0,0 +1,260 @@
...
+# Skip the entire module if the upstream pyvisa changes that NIENET100
+# depends on (InterfaceType.ni_enet100_tcpip + rname.NIEnet100TCPIPIntfc)
+# are not in place. The pyvisa_py.nienet100 module raises ImportError on
+# load when they are missing.
+try:
+    from pyvisa_py import nienet100 as _ni
+except ImportError as _import_err:
+    pytestmark = pytest.mark.skip(
+        reason="pyvisa-py NI GPIB-ENET/100 session layer unavailable: %s" % _import_err
+    )
```
MatthieuDartiailh: Same as in other places

MatthieuDartiailh (commented):
Some small comments but we are getting there. Regarding logging I think we can just drop the conditional. We can revisit later if it proves to be a problem.

pyvisa_py/nienet100.py
bytewarrior: Understood. We'll leave it as it is for now, but it can be changed when needed. Also, I think the comment above it summarizes the situation well enough.

pyvisa_py/nienet100.py
bytewarrior: This was fixed in commits 7f0a9ed and b15748b. The read timeout is now driven from the session timeout attribute rather than a fixed value: _sync_read_timeout() converts self.timeout into the bracket tmo_code and falls back to T10s (10 s) when the attribute is unset, and after_parsing seeds the operational socket timeout from self.timeout (else the connect default). It's applied both at open and on mid-session timeout changes via _set_timeout -> _sync_read_timeout.

pyvisa_py/testsuite/test_nienet100.py
bytewarrior: Applied.

bytewarrior: See above (duplicate?)

bytewarrior: See above (duplicate?)

pyvisa_py/protocols/nienet100.py
bytewarrior: The `if` line was removed with commit [db0ca25](https://github.com/pyvisa/pyvisa-py/pull/587/commits/db0ca2588b4c9b99f63b28b2810900e24fa86ce8).

pyvisa_py/testsuite/test_nienet100_session.py
bytewarrior: Done. pyvisa 1.17.0 (not yet confirmed, but assumed in the pyvisa repo) is now a dependency. Guard code has been removed. See commit 1aaf956.

pyvisa_py/testsuite/nienet100_assisted_tests/test_session.py
bytewarrior: Done, see above. Commit: 1aaf956

pyvisa_py/nienet100.py
bytewarrior: I have reworked the timeout machinery, this should be fixed now as well.

bytewarrior:
Hi Matthieu, I have to admit I've been busy and then there was a summer vacation... but I'm back now and I've seen there's a lot of activity in pyvisa-py. Please let me know, if I can help you with anything.

MatthieuDartiailh:
Thanks. I am trying to go through the last PRs to prepare the release. If you can test latest main that would be very helpful.

Issue #596, opened 2026-06-24
Read timeouts (EOI-related) when communicating with HP8903B audio analyzer


DavidCPlatt:
Communication with an HP8903B audio analyzer works only in one direction.  Writes work.  Reads always time out.

The HP8903B is an old device, and one of its quirks is that it does not assert EOI when sending the last byte of the response.  The GPIB adapter must terminate the read, based on the arrival of the termination character ('\n').

I have been testing with a Xyphro USB-to-GPIB interface, using both its native (USBTMC) and Linux-GPIB firmwares.

1. USBTMC, using pyvisa 1.16.2 and pyvisa_py 0.8.1, and the USBTMC firmware.  Reads always time out, at all settings of the termination character and the VISA attributes I have tried.
2. GPIB, using pyvisa 1.16.2 and pyvisa_py 0.8.1 and gpib_ctypes-0.3.0, and the Linux-GPIB firmware.   Reads always time out, at all settings of the termination character and the VISA attributes I have tried.
3. GPIB, using direct calls to gpib_ctypes-0.3.0, and the Linux-GPIB firmware.  Reads complete as they should, if I set the EOS character and terminate-read-on-EOS flags when the GPIB device is initialized.

Working with the Xyphro developer, and tracing USB activity and reading through the code, I've confirmed that in cases 1 and 2, the proper termination-character-and-mode settings are being ignored at a low level in the pyvisa_py code.  In the case of USBTMC, the termination character makes it down to the low-level read() method, but it's not being used when formatting the transfer attributes for the packet (the value of None is being passed to the bulk-out builder).  In the case of GPIB, the VISA attributes are set correctly in the GPIBSession, but the read() code doesn't look at them and make the necessary configuration calls to the lower-level GPIB device code.

Both of these shortfalls were rather easy to patch, and the patches work... all three of my test cases pass with the patches in place.

**Output of `pyvisa-info`**

Machine Details:
   Platform ID:    Linux-7.0.10+deb13-amd64-x86_64-with-glibc2.41
   Processor:      

Python:
   Implementation: CPython
   Executable:     /usr/src/pyvisa-updated/.venv/bin/python
   Version:        3.13.5
   Compiler:       GCC 14.2.0
   Architecture:   ('x86', 64)
   Build:          May  5 2026 21:05:52 (#main)
   Unicode:        UCS4

PyVISA Version: 1.16.2

Backends:
   ivi:
      Version: 1.16.2 (bundled with PyVISA)
[...]

PR #597, opened 2026-06-24
gpib: fix EOS for read

DavidCPlatt:
Add actual TERMCHAR support to GPIB interface
    
Use the TERMCHAR and TERM_CHAR_EN attributes to configure the GPIB interface prior to each read.  Partial fix for #596
    
This allows the GPIB device class to be used with devices such as the HP 8903B which do not signal EOI at the end of input and must depend on termination-character matching by the interface.

- [ X] Closes # 596 (partial fix)
- [ X] Added an entry to the CHANGES file

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,7 @@ PyVISA-py Changelog
   server to receive VXI-11 `DEVICE_INTR_SRQ` callbacks. This also fixes the
   `create_intr_chan` XDR packer in `protocols/vxi11.py`.
   Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
+- Pass termchar information to USBTMC and GPIO devices when reading.
 
 0.8.1 (04-09-2025)
 ------------------
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -467,14 +467,16 @@ class _GPIBCommon(Session):
 
         reader = lambda: ifc.read(count)  # noqa: E731
 
-        if (self.attrs[ResourceAttribute.termchar] is None
-            or not self.attrs[ResourceAttribute.termchar_enabled]):
-            ifc.config(0x0C, 0)         # IbcEOSrd
+        if (
+            self.attrs[ResourceAttribute.termchar] is None
+            or not self.attrs[ResourceAttribute.termchar_enabled]
+        ):
+            ifc.config(0x0C, 0)  # IbcEOSrd
         else:
             termchar = self.attrs[ResourceAttribute.termchar]
             ifc.config(0x0F, termchar)  # IbcEOSchar
-            ifc.config(0x0E, 1)         # IbcEOScmp
-            ifc.config(0x0C, 1)         # IbcEOSrd
+            ifc.config(0x0E, 1)  # IbcEOScmp
+            ifc.config(0x0C, 1)  # IbcEOSrd
 
         return self._read(reader, count, checker, False, None, False, gpib.GpibError)
 
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -467,6 +467,15 @@ class _GPIBCommon(Session):
 
         reader = lambda: ifc.read(count)  # noqa: E731
 
+        if (self.attrs[ResourceAttribute.termchar] is None
+            or not self.attrs[ResourceAttribute.termchar_enabled]):
+            ifc.config(0x0C, 0)         # IbcEOSrd
+        else:
+            termchar = self.attrs[ResourceAttribute.termchar]
+            ifc.config(0x0F, termchar)  # IbcEOSchar
+            ifc.config(0x0E, 1)         # IbcEOScmp
+            ifc.config(0x0C, 1)         # IbcEOSrd
+
         return self._read(reader, count, checker, False, None, False, gpib.GpibError)
 
     def write(self, data: bytes) -> Tuple[int, StatusCode]:


CHANGES
```diff
@@ -23,6 +23,7 @@ PyVISA-py Changelog
   server to receive VXI-11 `DEVICE_INTR_SRQ` callbacks. This also fixes the
   `create_intr_chan` XDR packer in `protocols/vxi11.py`.
   Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
+- Pass termchar information to USBTMC and GPIO devices when reading.
```
MatthieuDartiailh, line 26: Please mention the PR number

pyvisa_py/gpib.py
```diff
@@ -467,6 +467,17 @@ def read(self, count: int) -> Tuple[bytes, StatusCode]:
...
+        if (
+            self.attrs[ResourceAttribute.termchar] is None
+            or not self.attrs[ResourceAttribute.termchar_enabled]
+        ):
+            ifc.config(0x0C, 0)  # IbcEOSrd
+        else:
+            termchar = self.attrs[ResourceAttribute.termchar]
+            ifc.config(0x0F, termchar)  # IbcEOSchar
+            ifc.config(0x0E, 1)  # IbcEOScmp
+            ifc.config(0x0C, 1)  # IbcEOSrd
```
MatthieuDartiailh, line 479: Rather than doing this at each read, could it be possible to do it once in a handler for the attribute to avoid the systematic overhead.
Also if there not already an enum for the admissible values to config creating one would make sense.

CHANGES
DavidCPlatt: Please pardon my ignorance - how is that best done?   In the normal git process I have to commit the change to the file, before uploading to Gitlab, before creating the pull request (which is when I know what the pull-request number is).

Do I have to split changes into two (one commit and pull request for the actual change, and another for the documentation in CHANGES)?  or edit the change after the pull request is created, or ???.

Please educate me! :-)

pyvisa_py/gpib.py
DavidCPlatt: I will look at this.  From what I have seen of the Session and GPIOSession code structure, this would add some complications due to the fact that there are two interacting attributes - neither one alone controls what we must tell the interface to do.   I'd need to add several things to GPIOSession:  one actual "look at the two, and configure" method, and two custom setter functions for the two attributes which would set them and then call the common work function.  We would actually end up going through the configure process twice at a minimum (once each time one of the two attributes is set).  This would eliminate most of the overhead (I think) but at the expense of a good deal of additional code complexity.  Is there a leaner way, within the Attribute framework?

The other possibility would be to do the attribute lookup and comparison each time through read() as I do now, creating a tuple (actual or otherwise) of the three values for the three configuration settings.  The GPIOSession would store a cached value of the tuple.  On each read we'd compute the new tuple, compare it to the cached one, call ifc.config for each config item which has changed, and then update the cached tuple.  This would eliminate the ifc.config() calls unless something changed, at the expense of some significant data management each time.

The way I looked at it, was that the overhead in the "lazy" approach here ought to be quite small.  The configuration updates (changed or not) do go down to the interface driver, but they do not actually trigger any GPIO or USBTMC bus activity - the device itself isn't actually involved at this time.  That seemed a good tradeoff for keeping the change small and localized, and the impact on the GPIOSession code and data structure down.

So - what do you think?

DavidCPlatt: As far as the enums go... well, there's a nice set of them defined in gpib_ctypes/constants, but I wasn't sure whether pyvisa_py is formally dependent on gpib_ctypes - I saw something about installing either this _or_ a different GPIB package at one point.  I didn't want to risk introducing a dependency which would break things.  

If that's safe, I could work on another PR to import the GPIB constants, clean up this usage, and look for other places in the code where simple numeric values were used in place of a mnemonic, and fix those too.

CHANGES
MatthieuDartiailh: You can simply push the changelog change as a separate commit after creating the PR.

pyvisa_py/gpib.py
MatthieuDartiailh: Do you expect any kind of cross-talks between those attributes ? Because I do not understand why you would need to re-specify the term-char when enabling the term char.

If not not you should be able to simply manage those in `_get_attribute` and `_set_attribute` of the GPIB session.

Regarding the enum, we do not have a strong dependency on gpib-ctypes but we could duplicate the enum to clean up our code.

DavidCPlatt: Since the cross-talk is strictly an enable/disable issue, I think you're right.  I'll work up a different work-branch and set of changes along those lines, test it, and submit a PR with that approach and let you decide what (if anything) you want to approve.  Looking back, I think my approach was to keep the code interventions I was doing for GPIB and USBTMC in the same style (ensuring the correct read configuration as part of the read process).  That may be an unnecessary consistency, I suppose.

If it's OK with you I'll handle the "create or use a consistent set of enums" as a separate PR.  There are a lot of other uses of hard-coded constants-in-place-of-enums and I think it makes sense to clean them all up in a step, rather than pull them in here.

MatthieuDartiailh: Sounds good

DavidCPlatt: I'm going to abandon this PR, as the alternate implementation you suggested seems to have worked out very nicely and I think it will address your concerns.

I'll start working on the enum change.

MatthieuDartiailh:
Closing in favor of #closzd in favorbof #599

PR #600, opened 2026-06-26
Gpib enums

DavidCPlatt:
Clean up the GPIB code, by using symbolic constants (IntEnums) when calling the libgpib APIs, rather than bare integers.

Rather than trying to include the API constants directly out of gpib_ctypes (which would have introduced a strict dependency that doesn't seem to exist today) I copied and modified the constant definitions from that file, creating "gpib_constants.py".  The new constant definitions take the form of an individual IntEnum class per API-call, with the class name corresponding to the API call (or the returned-value interpretation) and the class member names matching the names used in gpib_ctypes.  This constants file is now imported by gpib.py.

In gpib.py, all of the bare-integer value uses (I think) are replaced by the corresponding symbol, and old commentary explaining the interpretation of the bare-integer values is removed as redundant.

Tested with my Xyphro adapter and HP2903B.  No errors encountered.

This deserves a careful set of eyeball checks before merging!  I have reviewed it a couple of times myself and think I've got all of the bare-integer-to-symbol correspondences correct, but you know how much an author's own code reviews are worth :-)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,8 @@ PyVISA-py Changelog
   Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
 - Pass termchar information to USBTMC devices when reading PR #598
 - Properly configure termchar on GPIB devices. This fix timeout issues for old GPIB devices which do not signal EOI PR #599
+- Clean up GPIB calls to the ib*() routines, by using symbolic definitions
+  rather than bare integers.  PR #600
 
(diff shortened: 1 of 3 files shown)


DavidCPlatt:
Ugh.  This seems to have picked up some unnecessary commits, from my local merges and rebases.  I'm going to abandon it, and cherry-pick the change into a fresh tree.

MatthieuDartiailh:
Thanks for doing that !!!

PR #601, opened 2026-06-26
Modify GPIB code to use symbolic constants

DavidCPlatt:
Create a file of symbolic constants (IntEnum classes) for the various libgpib interface calls.  Use these symbols rather than hard-coded integers.  Remove
now-extraneous commentary about what the hard-coded integers mean.

The new file gpib_constants.py is derived from the constants file in gpib_ctypes (which isn't a strict dependency of this project).  I've reorganized the values as an individual IntEnum class per API or purpose, with the class name based on the API or purpose, and the member names matching those used in gpib_ctypes.  This provides a clean visual match between the ib*() call, class name, and member name each time one of the symbols is used, and should help prevent usage mismatches.  In the gpib.py code I modified, I have removed the old commentary which explained the intent of each bare integer as being unnecessary.

This change deserves careful eyeball checking before being merged, to make sure I've gotten the translations between bare-integer and symbol correct in every case.  I've reviewed it all two or three times, but you know how much author code reviews are worth! :-)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,7 +25,8 @@ PyVISA-py Changelog
   Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
 - Pass termchar information to USBTMC devices when reading PR #598
 - Properly configure termchar on GPIB devices. This fix timeout issues for old GPIB devices which do not signal EOI PR #599
-
+- Clean up GPIB calls to the ib*() routines, by using symbolic definitions
+  rather than bare integers.  PR #601
 0.8.1 (04-09-2025)
 ------------------
 
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@ PyVISA-py Changelog
 - Properly configure termchar on GPIB devices. This fix timeout issues for old GPIB devices which do not signal EOI PR #599
 - Clean up GPIB calls to the ib*() routines, by using symbolic definitions
   rather than bare integers.  PR #601
+
 0.8.1 (04-09-2025)
 ------------------
 
--- a/pyvisa_py/gpib.py
+++ b/pyvisa_py/gpib.py
@@ -348,7 +348,7 @@ class _GPIBCommon(Session):
         minor = int(self.parsed.board)
         # Secondary address (SAD) values should be in the range 96 to 126,
         # 0 means the SAD is disabled.
-        sad = gpib_constants.sad.NO_SAD
+        sad = gpib_constants.sad.NO_SAD.value
         timeout = gpib_constants.timeout.T10s
         send_eoi = 1
         eos_mode = 0
@@ -356,7 +356,7 @@ class _GPIBCommon(Session):
         if isinstance(self.parsed, GPIBInstr):
             pad = int(self.parsed.primary_address)
             if self.parsed.secondary_address is not None:
-                sad = int(self.parsed.secondary_address) + int(gpib_constants.sad.FIRST_SAD)
+                sad = int(self.parsed.secondary_address) + gpib_constants.sad.FIRST_SAD
             # Used to talk to a specific resource
             self.interface = Gpib(
                 name=minor,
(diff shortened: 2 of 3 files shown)


pyvisa_py/gpib.py
```diff
@@ -672,63 +659,58 @@ def _set_attribute(
...
                 return StatusCode.error_nonsupported_attribute_state
 
         elif attribute == ResourceAttribute.gpib_secondary_address:
-            # IbcSAD 0x2
             # Add 0x60 because National Instruments.
             if isinstance(attribute_state, int) and 0 <= attribute_state <= 30:
-                if ifc.ask(2):
-                    ifc.config(2, attribute_state + 96)
+                if ifc.ask(gpib_constants.ask.IbaSAD):
+                    ifc.config(gpib_constants.config.IbcSAD, attribute_state + 96)
```
MatthieuDartiailh: I wish my comment about the 96 was more explicit...

MatthieuDartiailh (commented):
Looks fine to me but I will do another pass later.

pyvisa_py/gpib.py
DavidCPlatt: Ach.  I didn't even consider those NI-specific SAD masks/constants.

I'll add another commit to the PR for these.  Shouldn't take long.

DavidCPlatt: Things being as they are, these days, I'm not absolutely certain that saying "96 86'ed" is entirely safe, but what the hell.  Done.

DavidCPlatt: Hah! Just realized there was an off-by-one error in the old code - range(96,126) should have been range(96,127).  As it was, subaddress 30 wouldn't have been iterated.  Fixed now.

CHANGES
```diff
@@ -25,7 +25,8 @@ PyVISA-py Changelog
   Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
 - Pass termchar information to USBTMC devices when reading PR #598
 - Properly configure termchar on GPIB devices. This fix timeout issues for old GPIB devices which do not signal EOI PR #599
-
+- Clean up GPIB calls to the ib*() routines, by using symbolic definitions
+  rather than bare integers.  PR #601
```
MatthieuDartiailh, line 29:
```suggestion
  rather than bare integers.  PR #601

```

pyvisa_py/gpib.py
```diff
@@ -205,21 +210,17 @@ def _analyse_lines_value(value: int, line: int):
 
     """
     if line == constants.VI_ATTR_GPIB_REN_STATE:
-        # REN bit valid = 0x10, REN bit value = 0x100
-        validity_mask = 0x10
-        value_mask = 0x100
```
MatthieuDartiailh, line 210: The constant is 0x1000, I assume this is actually fixing a bug.

MatthieuDartiailh (commented):
LGTM (minus the changelog formatting)

pyvisa_py/gpib.py
DavidCPlatt: I think so - good catch!

I just checked this against gpib_ctypes, and against the linux-gpib-4.3.7 uapi/gpib.h file.  In both of those, the value mask for REN is 0x1000 - a value mask of 0x0100 is for DAV and corresponds to a "DAV valid" bit of 0x01.

So, yes, I think this is another latent bug being fixed... a slip-of-the keyboard at some point, quite possibly.

MatthieuDartiailh:
Can you address the linting/typin failure ? (mostly a bad import I think)

DavidCPlatt:
OK, this looks better.

I fixed the two enum-name-change sets (GPIBconfig and GPIBerror) I missed during my original restructuring.

I've moved the import into the proper sorted position.  There's one other complaint about an old import, but I don't know if I should touch it now (I didn't, before).

I fixed the type mismatch involving the SAD, although it took a couple of tries for me to realize what was actually going on.  Apparently IntEnums don't always behave precisely like integers, from a type-checking perspective.

Please take another look.

MatthieuDartiailh:
Looks good. One more import sorting you can use ruff to fix.

DavidCPlatt:
Done.

PR #606, opened 2026-07-23
vxi-11: doing readstb inside srq handler is not standard and potentially breaking

### LINKED ISSUE #603: Difference in SRQ handling between other IVI backends and pyvisa-py
this is in reference to PR #577

NI-VISA, in [Status/Service Request Service](https://www.ni.com/docs/en-US/bundle/ni-visa/page/statusservice-request-service.html) says the following about SRQ handling:

> Under VISA, the VI_EVENT_SERVICE_REQ event contains no additional information other than the type of event. Therefore, by using viGetAttribute() on the eventContext parameter, as shown in the following code, the program can identify the event as a service request.
`status = viGetAttribute(eventContext,VI_ATTR_EVENT_TYPE, &eventType);`
You can retrieve the status byte of the device by issuing a viReadSTB() operation. This is especially important because on some interfaces, such as GPIB, it is not always possible to know which device has asserted the service request until a viReadSTB() is performed. This means that all sessions to devices on the bus with the service request may receive a service request event. Therefore, you should always check the status byte to ensure that your device was the one that requested service. Even if you have only one device asserting a service request, you should still call viReadSTB() to guarantee delivery of future service request events.

(Haven't yet looked at what other VISA providers say, but NI-VISA is rather widely used.)

So: when you get an interrupt, they say it is up to the client to do `read_stb()`, and check for bit 6 (`0x40`), as you might get false positives. 

Note that this is not needed with gateways that filter events, like the ICS-9065, as it already does the read_stb, but it is needed with most other gateways, like the E5810. (In order to keep compatibility with the above 'guidelines', the gateways that I have seen that do filter SRQ, temporarily retain bit 6, and return it upon the next `read_stb()` call.)

**The subject:** 

pyvisa-py filters the events: does the `read_stb()` on itself upon reception of the interrupt call, but does NOT return bit 6 upon the next `read_stb()` call.

None of the IVI backends I've tested (NI-Visa, Keysight, R&S) do that `read_stb()`, they just forward the event to the client.

**The problems:**

* It is non standard behaviour, and makes code that works with the other IVI backends, fail.
* It is not (yet) documented
* Some older instruments react badly to the `read_stb()`. 

My recommendations would be:

1) allow disabling the automatic `read_stb()`
2) make it explicit in the documentation that pyvisa-py does (or can do) `read_stb()` (The documentation needs updating anyway, as it hasn't been updated since PR #577)
3) if pyvisa-py does `read_stb()` by itself, it should cache bit 6, and serve it back upon the next `read_stb()` call. 

I'm not saying that it needs to be removed, but it should be possible to make it compliant with expectations.

hb020:
- [x] Closes #603
       There was no documentation of this behaviour

Reasoning is mentioned in #603. The used approach is to remove the `read_stb()`.  The only other feasible approach would require 2 things:
- Add a setting to select enabling it or not. 
- Add a cache of the SRQ bit, like the ICS-9065 does. 

This would however be highly specific to pyvisa-py, for a seldomly used feature, would require a lot of work, and would potentially confuse people. Not worth it.

**Details of the why**, with rewording of the referred issue:

This is about the SRQ handling for VXI-11, where a device can raise the `SRQ` state, and sends an event to the VXI-11 client. 

For connections to single devices, there is no confusion as to who provoked the event. 

But when you connect to a VXI-11.2 gateway (E5810, ICS-9065, and others), the `SRQ` may come from any device on the GPIB bus, as there is only 1 line for the entire bus. One device may have raised the event, but all clients that are connected to any device on the bus (and have event handling activated), will receive the event. **ALL** IVI backends I have tested (NI-Visa, Keysight, R&S) therefore expect the client to then to actively determine which device provoked that event, as you may get false positives. The VXI-11.2 specification mentions this behaviour (Rule B.4.13), and NI-Visa even explicitly mentions the requirement for clients to filter events themselves. 

Determining who really raised the event is easy with IEEE-488.2 compatible devices (using `read_stb()`), but IEEE-488.1 or earlier devices may not support that same method. Some may even freak out over it. So: filtering of the events by some middle man (meaning: pyvisa) is not only non standard, but even potentially breaking on some instruments. The implementation that was in place did the `read_stb()` 'hard wired', without possibility to switch it off.

The implementation that was in place was even more breaking, as once you call `read_stb()`, you 'consume' the SRQ bit, and it is no longer there. The client that then wants to call `read_stb()` to determine if the device really provoked the event (like how it is expected on the other IVI backends), will no longer find any proof of it, and reject the event. So in order to keep compatibility with the other IVI backends, you need the cache that is mentioned above.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,9 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling, as it is not standard behaviour, may confuse instruments, and does not cache the SRQ bit as would be required. Closes #603
+- Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling. 
+  It is not standard behaviour, may confuse instruments, and does not 
+  cache the SRQ bit as would be required. Closes #603 PR #606
 - Added read_stb, assert_trigger and gpib_control_run functions to usbtmc PR #532
 - Implement viTerminate() for HiSLIP sessions using a CancellableSocket
   (self-pipe trick on recv_into), enabling cross-thread I/O cancellation
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling, as it is not standard behaviour, may confuse instruments, and does not cache the SRQ bit as would be required. Closes #603
 - Added read_stb, assert_trigger and gpib_control_run functions to usbtmc PR #532
 - Implement viTerminate() for HiSLIP sessions using a CancellableSocket
   (self-pipe trick on recv_into), enabling cross-thread I/O cancellation
--- a/pyvisa_py/events.py
+++ b/pyvisa_py/events.py
@@ -46,7 +46,6 @@ class EventContext:
     """Immutable description of a single VISA event occurrence."""
 
     event_type: constants.EventType
-    status_byte: int = 0
     timestamp: float = field(default_factory=time.time)
     context_id: int = field(default_factory=lambda: random.getrandbits(32))
 
--- a/pyvisa_py/protocols/vxi11.py
+++ b/pyvisa_py/protocols/vxi11.py
@@ -498,12 +498,9 @@ class SrqInterruptTCPServer(rpc.TCPServer):
             # Defensive: session may have been closed while we were spawned
             if self.session.interface is None or self.session.link == 0:
                 return
-            stb, status = self.session.read_stb()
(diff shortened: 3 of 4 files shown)


CHANGES
```diff
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling, as it is not standard behaviour, may confuse instruments, and does not cache the SRQ bit as would be required. Closes #603
```
```suggestion
- Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling. 
  It is not standard behaviour, may confuse instruments, and does not 
  cache the SRQ bit as would be required. Closes #603 PR #606
```

MatthieuDartiailh:
Minor comment regarding the changelog but this can be merged once addressed.

Thanks for your contribution

MatthieuDartiailh:
CI failure is unrelated

PR #612, opened 2026-08-07
[vxi-11] read function bug corrections 

### LINKED ISSUE #608: VXI-11: END/EOI is discarded when response length is an exact multiple of chunk_size, causing VI_ERROR_TMO
## Summary

When the total length of a VXI-11 response is an exact multiple of `chunk_size`,
`TCPIPInstrVxi11.read()` returns `StatusCode.success_max_count_read` even though the
instrument set the END/EOI bit in `reason`. `MessageBasedResource._read_raw()` interprets
that status as "more data pending" and issues another `device_read`. The instrument has
nothing left to send, so the call blocks until the VISA timeout expires and raises
`VI_ERROR_TMO`.

The instrument behaves correctly: it signals END together with the last data block. The
information is received by pyvisa-py but never evaluated.

This is fully deterministic. With a response length that is an exact multiple of
`chunk_size` it fails every time; one byte more and it succeeds every time.

## Root cause

`pyvisa_py/tcpip.py`, `TCPIPInstrVxi11.read()` (current `main`):

```python
while reason & end_reason == 0:              # exit 1: checks reason
    chunk_timeout = max(10, timeout - int((time.time() - start_time) * 1000))
    error, reason, data = read_fun(          # reason is populated here
        self.link, chunk_length, chunk_timeout,
        self.lock_timeout, flags, term_char,
    )
    if error == vxi11.ErrorCodes.io_timeout:
        return bytes(read_data), StatusCode.error_timeout
    elif error:
        return bytes(read_data), StatusCode.error_io
    read_data.extend(data)
    count -= len(data)
    if count <= 0:                           # exit 2: checks count only
        status = StatusCode.success_max_count_read
        break
    chunk_length = min(count, chunk_length)
return bytes(read_data), status
```

The loop has two exits, and only one of them looks at `reason`:

* Exit 1 is the `while` condition, evaluated only at the top of an iteration.
* Exit 2 is the `break`, taken from inside the loop body.

When `count` reaches exactly zero, the `break` is taken and the `while` condition is never
re-evaluated. At that moment `reason` holds the END bit, set a few lines earlier, but
nothing reads it. Every exit through the `break` is therefore blind to END, regardless of
whether it was set.

The counter check itself is correct and necessary — the caller asked for at most
`chunk_size` bytes. What is missing is that the END information is not propagated when the
loop exits this way. There is no status distinguishing "buffer full, more pending" from
"buffer full and complete".

## Steps to reproduce

The response length is forced onto the boundary via `chunk_size`, so no particular data
size or instrument-specific payload is required.

```python
import pyvisa

inst = rm.open_resource('TCPIP::<address>::INSTR')
inst.timeout = 3000

# 1) Measure the exact response length (large chunk_size rules out the bug)
inst.chunk_size = 10 * 1024 * 1024
inst.write('<query returning at least a few kB>')
L = len(inst.read_raw())
print('length:', L)

# 2) Set chunk_size to exactly that length
inst.chunk_size = L
inst.write('<same query>')
inst.read_raw()          # -> VI_ERROR_TMO, reproducibly

# 3) One byte more
inst.chunk_size = L + 1
inst.write('<same query>')
inst.read_raw()          # -> succeeds, reproducibly
```

Steps 2 and 3 were verified on the setup described below; both are reproducible on demand.

Reading the code, `chunk_size = L - 1` should also succeed: the counter reaches zero, but
one byte genuinely remains and arrives in the next iteration with END set. That is the case
the counter check is meant for. I have not verified this variant.

## Expected vs. actual behaviour

**Expected:** the read completes and returns the data, because the instrument signalled END.

**Actual:** `VI_ERROR_TMO` after the full VISA timeout.

## Wireshark evidence

Captured on a failing read (VXI-11 over TCP, VISA timeout 140 s):

* The final `DEVICE_READ Reply` carries `Reason: 0x00000005` — `Requested Count Reached: True`
  and `EOI Set: True`. Both statements are true here: the chunk was filled exactly and it
  was the end of the message. Multiple reason bits at once are permitted by the VXI-11
  specification.
* A successful read of the same query shows `Requested Count Reached: False`, `EOI Set: True`
  on its final reply.
* Approximately 200 µs after the final data reply, pyvisa-py issues another `DEVICE_READ`
  with no preceding `DEVICE_WRITE`.
* The instrument answers that read with `I/O Timeout`. No retransmissions or duplicate ACKs
  — the connection is healthy, there is simply nothing left to send.
* Because the elapsed time up to that point is negligible, `chunk_timeout` grants the
  superfluous read almost the entire configured timeout. In my capture the interval between
  the extra `DEVICE_READ` and the `I/O Timeout` reply was 139.82 s.

## Suggested fix

Evaluate `reason` when leaving through the `break`:

```python
    if count <= 0:
        if reason & end_reason:
            status = StatusCode.success
        else:
            status = StatusCode.success_max_count_read
        break
```

This is untested — I have not run it or the test suite against this change.

## Notes

* Possibly related: #101 — same counter in the same function, opposite edge case (counter
  going far negative when the device returns more than requested). That report also ends
  with an "increase the requested size" workaround rather than a diagnosis.
* The PyVISA documentation has long suggested changing `chunk_size` when a message cannot
  be read completely, while stating that the underlying mechanism is unclear. This may be
  that mechanism.
* Workaround: set `chunk_size` larger than the largest possible response, so the counter can
  never reach zero.
* ASCII transfers make this appear sporadic, because the response length varies between
  queries and only occasionally lands on a multiple of `chunk_size`. This is how I ran into
  it: a long-running measurement that failed after one to four hours at random. With
  fixed-length binary transfers the outcome is deterministic instead — either always fine or
  always broken, depending on the number of points.

**Output of `pyvisa-info`**

```
Machine Details:
   Platform ID:    Windows-11-10.0.26200-SP0
   Processor:      Intel64 Family 6 Model 165 Stepping 3, GenuineIntel

Python:
   Implementation: CPython
   Executable:     C:\Users\xxxyyyxxx\anaconda3\python.exe
   Version:        3.12.7
   Compiler:       MSC v.1929 64 bit (AMD64)
   Architecture:   AMD64
   Build:          Oct  4 2024 13:17:27 (#main)
   Unicode:        UCS4

PyVISA Version: 1.15.0

Backends:
   ivi:
      Version: 1.15.0 (bundled with PyVISA)
      #1: C:\WINDOWS\system32\visa32.dll:
         found by: auto
         architecture:
            ('x86', 64)
         Vendor: National Instruments
         Impl. Version: 22021376
         Spec. Version: 5244928
      #2: C:\WINDOWS\system32\visa64.dll:
         found by: auto
         architecture:
            ('x86', 64)
         Vendor: National Instruments
         Impl. Version: 22021376
         Spec. Version: 5244928
   py:
      Version: 0.8.0
      TCPIP INSTR: Available
         Resource discovery:
         - VXI-11: ok
      TCPIP SOCKET: Available
      PRLGX_TCPIP INTFC: Available
      GPIB INSTR: Available
      ASRL INSTR:
         Please install PySerial (>=3.0) to use this resource type.
         No module named 'serial'
      USB INSTR:
         No module named 'usb'
      USB RAW:
         No module named 'usb'
      VICP INSTR:
      PRLGX_ASRL INTFC:
         Please install PySerial (>=3.0) to use this resource type.
         No module named 'serial'
      GPIB INTFC:
         Please install linux-gpib (Linux) or gpib-ctypes (Windows, Linux) to use this resource type. Note that installing gpib-ctypes will give you access to a broader range of functionalities.
         No module named 'gpib'
```

hb020:
**Why**

* On certain `inst.chunk_size` values in relation to the content size, a read falls in timeout. This is corrected.
* Read function:
    * return values (status codes) were incomplete. It is now compliant, and can return all 3 non-error possibilities: `success`, `success_termination_character_read`, `success_max_count_read`, all while taking into account `VI_ATTR_SUPPRESS_END_EN`
    * count boundary check (was absent)
    * timeout handling (was incomplete)
    * zero length handling (was incomplete)

**Tasks**

- [x] Closes #608 (and corrected some spacing problems and removed some trailing spaces from CHANGES.md)
- [ ] Executed ``black . && isort -c . && flake8`` with no errors. -> Is this still valid? It seems outdated.
- [x] The change is fully covered by automated unit tests.  Well, not really. Could do it, but will take a lot of time as the framework for this type of tests is lacking in my setup. (need the keysight tools, likely have to find a Windows machine....) But see below for a test setup.
- [ ] Documented in docs/ as appropriate. -> There was no doc of this bug. I might create another PR to document the gaps between pyvisa-py and the 'de facto standard' NI-Visa in more detail.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,10 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- Corrected timeout caused by when response length is an exact multiple of chunk_size
+- Read function termination overhaul:
+  Corrected timeout in read when response length is an exact multiple
+  of chunk_size, implemented read count boundary checks, completed termination
+  character detection, allowed use of suppress_end_en, corrected timeout handling.
   Closes #608 PR #612
 - Removed return packet on device_intr_srq. Closes #614 PR #615
 - Removed SRQ server shutdown if the server was already shut down,
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,10 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- Removed SRQ server shutdown if the server was already shut down, reducing potential problems with non compliant devices. Closes #609 PR #610
+- Corrected timeout caused by when response length is an exact multiple of chunk_size
+  Closes #608 PR #612
+- Removed SRQ server shutdown if the server was already shut down,
+  reducing potential problems with non compliant devices. Closes #609 PR #610
 - Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling.
   It is not standard behaviour, may confuse instruments, and does not
   cache the SRQ bit as would be required. Closes #603 PR #606
@@ -13,11 +16,11 @@ PyVISA-py Changelog
   (self-pipe trick on recv_into), enabling cross-thread I/O cancellation
   without destroying the session. Also handle HiSLIP Interrupted messages
   in the receive path. Closes #566
-- Fixed VXI-11 read timeouthandling. When reads were splitdown into multiple read 
-  chunks the timeout value for each chunk was calculated wrongly. This resulted in 
+- Fixed VXI-11 read timeout handling. When reads were split down into multiple read
+  chunks the timeout value for each chunk was calculated wrongly. This resulted in
   very high timeout values which had to be set.
-  A second addressed issue is that timeout values never decrement to 0. A timeout value 
-  of 0 is undefined in VXI-11 standard. It can mean "timeout immediately if no data 
+  A second addressed issue is that timeout values never decrement to 0. A timeout value
+  of 0 is undefined in VXI-11 standard. It can mean "timeout immediately if no data
   is in buffer" or "block permanently until transfer is finished".
 - Implement the VISA event subsystem for VXI-11 (TCPIP::INSTR) resources:
   `viEnableEvent`, `viDisableEvent`, `viDiscardEvents`, `viWaitOnEvent`,
@@ -26,7 +29,7 @@ PyVISA-py Changelog
   (`install_handler`) delivery. A daemon thread runs an ONC RPC TCP interrupt
   server to receive VXI-11 `DEVICE_INTR_SRQ` callbacks. This also fixes the
   `create_intr_chan` XDR packer in `protocols/vxi11.py`.
-  Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577 
+  Other transports (GPIB, USBTMC, HiSLIP, Serial) remain unsupported for now. PR #577
 - Pass termchar information to USBTMC devices when reading PR #598
 - Properly configure termchar on GPIB devices. This fix timeout issues for old GPIB devices which do not signal EOI PR #599
 - Clean up GPIB calls to the ib*() routines, by using symbolic definitions
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- Read function termination overhaul:
+- VXI-11 (TCPIP::INSTR) Read function termination overhaul:
   Corrected timeout in read when response length is an exact multiple
   of chunk_size, implemented read count boundary checks, completed termination
   character detection, allowed use of suppress_end_en, corrected timeout handling.
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -741,7 +741,7 @@ class TCPIPInstrVxi11(Session):
         # Get the timeout as cleaned up by the upper layers
         timeout = self._io_timeout
 
-        # See if a timeout was really set. 
+        # See if a timeout was really set.
         # if self.timeout is None, the given timeout was VI_TMO_INFINITE
         # This lookup method is also slightly convoluted because the unit test scripts
         # may not have set this correctly.
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -752,7 +752,10 @@ class TCPIPInstrVxi11(Session):
             count -= len(data)
 
             if count <= 0:
-                status = StatusCode.success_max_count_read
+                if reason & end_reason:
+                    status = StatusCode.success
+                else:
+                    status = StatusCode.success_max_count_read
                 break
 
             chunk_length = min(count, chunk_length)
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -738,9 +738,13 @@ class TCPIPInstrVxi11(Session):
         read_fun = self.interface.device_read
         status = StatusCode.success
 
-        timeout = self._io_timeout  # this is derived from from self.timeout
-        # See if a timeout was set. This is derived from self.timeout, but slightly
-        # convoluted because the unit test scripts may not have set this correctly.
+        # get the timeout as cleaned up by the upper layers
+        timeout = self._io_timeout
+        
+        # See if a timeout was really set. This is done by looking at the timeout
+        # that was given to the interface. If it is None, then there is no timeout. 
+        # This lookup method is also slightly convoluted because the unit test scripts
+        # may not have set this correctly.
         finite_timeout = getattr(self, "timeout", None) is not None
         start_time = time.time()
         while reason & stop_reason == 0:
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -759,10 +759,13 @@ class TCPIPInstrVxi11(Session):
             # Calculate timeout for current chunk.
             # Limit the minimum timeout to 10ms, because 0 is undefined
             # according to VXI-11 spec (done also by Visas)
-            elapsed_ms = int((time.time() - start_time) * 1000)
-            remaining_timeout = timeout - elapsed_ms
-            if finite_timeout and remaining_timeout <= 0:
-                return bytes(read_data), StatusCode.error_timeout
+            if finite_timeout:
+                elapsed_ms = int((time.time() - start_time) * 1000)
+                remaining_timeout = timeout - elapsed_ms
+                if remaining_timeout <= 0:
+                    return bytes(read_data), StatusCode.error_timeout
+            else:
+                remaining_timeout = 2**32 - 1  # VI_TMO_INFINITE
             chunk_timeout = max(10, remaining_timeout)
             error, reason, data = read_fun(
                 self.link,
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -740,9 +740,9 @@ class TCPIPInstrVxi11(Session):
         status = StatusCode.success
 
         timeout = self._io_timeout  # this is derived from from self.timeout
-        # See if a timeout was set. This is derived from self.timeout, but slightly 
-        # convoluted because the unit test scripts may not have set this correctly.      
-        finite_timeout = getattr(self, "timeout", None) is not None  
+        # See if a timeout was set. This is derived from self.timeout, but slightly
+        # convoluted because the unit test scripts may not have set this correctly.
+        finite_timeout = getattr(self, "timeout", None) is not None
         start_time = time.time()
         while reason & stop_reason == 0:
             # Decrease timeout so that the total timeout does not get larger
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -744,7 +744,7 @@ class TCPIPInstrVxi11(Session):
         # See if a timeout was really set.
         # if self.timeout is None, the given timeout was VI_TMO_INFINITE
         finite_timeout = self.timeout is not None
-        
+
         start_time = time.time()
         while reason & stop_reason == 0:
             # Decrease timeout so that the total timeout does not get larger
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -743,9 +743,8 @@ class TCPIPInstrVxi11(Session):
 
         # See if a timeout was really set.
         # if self.timeout is None, the given timeout was VI_TMO_INFINITE
-        # This lookup method is also slightly convoluted because the unit test scripts
-        # may not have set this correctly.
-        finite_timeout = getattr(self, "timeout", None) is not None
+        finite_timeout = self.timeout is not None
+        
         start_time = time.time()
         while reason & stop_reason == 0:
             # Decrease timeout so that the total timeout does not get larger
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -740,9 +740,9 @@ class TCPIPInstrVxi11(Session):
 
         # get the timeout as cleaned up by the upper layers
         timeout = self._io_timeout
-        
+
         # See if a timeout was really set. This is done by looking at the timeout
-        # that was given to the interface. If it is None, then there is no timeout. 
+        # that was given to the interface. If it is None, then there is no timeout.
         # This lookup method is also slightly convoluted because the unit test scripts
         # may not have set this correctly.
         finite_timeout = getattr(self, "timeout", None) is not None
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -765,7 +765,7 @@ class TCPIPInstrVxi11(Session):
                 if remaining_timeout <= 0:
                     return bytes(read_data), StatusCode.error_timeout
             else:
-                remaining_timeout = 2**32 - 1  # VI_TMO_INFINITE
+                remaining_timeout = constants.VI_TMO_INFINITE  # This is 2**32 - 1
             chunk_timeout = max(10, remaining_timeout)
             error, reason, data = read_fun(
                 self.link,
@@ -1027,15 +1027,15 @@ class TCPIPInstrVxi11(Session):
 
     def _set_timeout(self, attribute: ResourceAttribute, value: int) -> StatusCode:
         """Sets timeout calculated value from python way to VI_ way"""
+        # value is in milliseconds,
+        # and can be VI_TMO_INFINITE (2**32 - 1) or VI_TMO_IMMEDIATE (0)
+        self._io_timeout = value
         if value == constants.VI_TMO_INFINITE:
             self.timeout = None
-            self._io_timeout = 2**32 - 1
         elif value == constants.VI_TMO_IMMEDIATE:
             self.timeout = 0
-            self._io_timeout = 0
         else:
             self.timeout = value / 1000.0
-            self._io_timeout = int(self.timeout * 1000)
         return StatusCode.success
 
 
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -738,11 +738,11 @@ class TCPIPInstrVxi11(Session):
         read_fun = self.interface.device_read
         status = StatusCode.success
 
-        # get the timeout as cleaned up by the upper layers
+        # Get the timeout as cleaned up by the upper layers
         timeout = self._io_timeout
 
-        # See if a timeout was really set. This is done by looking at the timeout
-        # that was given to the interface. If it is None, then there is no timeout.
+        # See if a timeout was really set. 
+        # if self.timeout is None, the given timeout was VI_TMO_INFINITE
         # This lookup method is also slightly convoluted because the unit test scripts
         # may not have set this correctly.
         finite_timeout = getattr(self, "timeout", None) is not None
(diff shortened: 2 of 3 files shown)


hb020:
About the tests, if it's worth something:

test code:
```python
import pyvisa

def testcase(provider, size: int):
    if provider == '@py':
        print(f"******* provider = 'pyvisa-py', data size = {size} *******")
    else:
        print(f"******* provider = 'NI-VISA', data size = {size} *******")
    rm = pyvisa.ResourceManager(provider)
    inst = rm.open_resource('TCPIP::192.168.7.116::inst5::INSTR')
    inst.timeout = 3000
    data_size = size - 2  # add the space for CR/LF
    myquery = f"longrd? {data_size}"  # this is query to a custom device that can return any length of data, up to 2^32-1 bytes. The data_size does not take into account the CR/LF

    # Measure the exact response length (large chunk_size rules out the bug)
    inst.chunk_size = 10 * 1024 * 1024
    inst.write(myquery)
    L = len(inst.read_raw())
    
    for chunk_size in range(L-9, L+11):
        # do a series of number of bytes less and more
        inst.chunk_size = chunk_size
        inst.write(myquery)
        try:
            inst.read_raw()          # -> fails, reproducibly, initially
        except Exception as e:
            print(f'read_raw() with chunk_size {inst.chunk_size} failed: {e}')
            continue
        print(f'read_raw() with chunk_size {inst.chunk_size} succeeded')
    
    
if __name__ == '__main__':
    testcase('@py', 10)
    testcase('@py', 2000)
    testcase('', 10)
    testcase('', 2000)
```

Without repair:
```text
python3 issue608test.py
******* provider = 'pyvisa-py', data size = 10 *******
read_raw() with chunk_size 1 failed: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
read_raw() with chunk_size 2 failed: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
read_raw() with chunk_size 3 succeeded
read_raw() with chunk_size 4 succeeded
read_raw() with chunk_size 5 failed: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
read_raw() with chunk_size 6 succeeded
read_raw() with chunk_size 7 succeeded
read_raw() with chunk_size 8 succeeded
read_raw() with chunk_size 9 succeeded
read_raw() with chunk_size 10 failed: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
read_raw() with chunk_size 11 succeeded
read_raw() with chunk_size 12 succeeded
read_raw() with chunk_size 13 succeeded
read_raw() with chunk_size 14 succeeded
read_raw() with chunk_size 15 succeeded
read_raw() with chunk_size 16 succeeded
read_raw() with chunk_size 17 succeeded
read_raw() with chunk_size 18 succeeded
read_raw() with chunk_size 19 succeeded
read_raw() with chunk_size 20 succeeded
******* provider = 'pyvisa-py', data size = 2000 *******
read_raw() with chunk_size 1991 succeeded
read_raw() with chunk_size 1992 succeeded
read_raw() with chunk_size 1993 succeeded
read_raw() with chunk_size 1994 succeeded
read_raw() with chunk_size 1995 succeeded
read_raw() with chunk_size 1996 succeeded
read_raw() with chunk_size 1997 succeeded
read_raw() with chunk_size 1998 succeeded
read_raw() with chunk_size 1999 succeeded
read_raw() with chunk_size 2000 failed: VI_ERROR_TMO (-1073807339): Timeout expired before operation completed.
read_raw() with chunk_size 2001 succeeded
read_raw() with chunk_size 2002 succeeded
read_raw() with chunk_size 2003 succeeded
read_raw() with chunk_size 2004 succeeded
read_raw() with chunk_size 2005 succeeded
read_raw() with chunk_size 2006 succeeded
read_raw() with chunk_size 2007 succeeded
read_raw() with chunk_size 2008 succeeded
read_raw() with chunk_size 2009 succeeded
read_raw() with chunk_size 2010 succeeded
******* provider = 'NI-VISA', data size = 10 *******
read_raw() with chunk_size 1 succeeded
read_raw() with chunk_size 2 succeeded
read_raw() with chunk_size 3 succeeded
read_raw() with chunk_size 4 succeeded
read_raw() with chunk_size 5 succeeded
read_raw() with chunk_size 6 succeeded
read_raw() with chunk_size 7 succeeded
read_raw() with chunk_size 8 succeeded
read_raw() with chunk_size 9 succeeded
read_raw() with chunk_size 10 succeeded
read_raw() with chunk_size 11 succeeded
read_raw() with chunk_size 12 succeeded
read_raw() with chunk_size 13 succeeded
read_raw() with chunk_size 14 succeeded
read_raw() with chunk_size 15 succeeded
read_raw() with chunk_size 16 succeeded
read_raw() with chunk_size 17 succeeded
read_raw() with chunk_size 18 succeeded
read_raw() with chunk_size 19 succeeded
read_raw() with chunk_size 20 succeeded
******* provider = 'NI-VISA', data size = 2000 *******
read_raw() with chunk_size 1991 succeeded
read_raw() with chunk_size 1992 succeeded
read_raw() with chunk_size 1993 succeeded
read_raw() with chunk_size 1994 succeeded
read_raw() with chunk_size 1995 succeeded
read_raw() with chunk_size 1996 succeeded
read_raw() with chunk_size 1997 succeeded
read_raw() with chunk_size 1998 succeeded
read_raw() with chunk_size 1999 succeeded
read_raw() with chunk_size 2000 succeeded
read_raw() with chunk_size 2001 succeeded
read_raw() with chunk_size 2002 succeeded
read_raw() with chunk_size 2003 succeeded
read_raw() with chunk_size 2004 succeeded
read_raw() with chunk_size 2005 succeeded
read_raw() with chunk_size 2006 succeeded
read_raw() with chunk_size 2007 succeeded
read_raw() with chunk_size 2008 succeeded
read_raw() with chunk_size 2009 succeeded
read_raw() with chunk_size 2010 succeeded
```

With repair:
```text
python3 issue608test.py              
******* provider = 'pyvisa-py', data size = 10 *******
read_raw() with chunk_size 1 succeeded
read_raw() with chunk_size 2 succeeded
read_raw() with chunk_size 3 succeeded
read_raw() with chunk_size 4 succeeded
read_raw() with chunk_size 5 succeeded
read_raw() with chunk_size 6 succeeded
read_raw() with chunk_size 7 succeeded
read_raw() with chunk_size 8 succeeded
read_raw() with chunk_size 9 succeeded
read_raw() with chunk_size 10 succeeded
read_raw() with chunk_size 11 succeeded
read_raw() with chunk_size 12 succeeded
read_raw() with chunk_size 13 succeeded
read_raw() with chunk_size 14 succeeded
read_raw() with chunk_size 15 succeeded
read_raw() with chunk_size 16 succeeded
read_raw() with chunk_size 17 succeeded
read_raw() with chunk_size 18 succeeded
read_raw() with chunk_size 19 succeeded
read_raw() with chunk_size 20 succeeded
******* provider = 'pyvisa-py', data size = 2000 *******
read_raw() with chunk_size 1991 succeeded
read_raw() with chunk_size 1992 succeeded
read_raw() with chunk_size 1993 succeeded
read_raw() with chunk_size 1994 succeeded
read_raw() with chunk_size 1995 succeeded
read_raw() with chunk_size 1996 succeeded
read_raw() with chunk_size 1997 succeeded
read_raw() with chunk_size 1998 succeeded
read_raw() with chunk_size 1999 succeeded
read_raw() with chunk_size 2000 succeeded
read_raw() with chunk_size 2001 succeeded
read_raw() with chunk_size 2002 succeeded
read_raw() with chunk_size 2003 succeeded
read_raw() with chunk_size 2004 succeeded
read_raw() with chunk_size 2005 succeeded
read_raw() with chunk_size 2006 succeeded
read_raw() with chunk_size 2007 succeeded
read_raw() with chunk_size 2008 succeeded
read_raw() with chunk_size 2009 succeeded
read_raw() with chunk_size 2010 succeeded
******* provider = 'NI-VISA', data size = 10 *******
read_raw() with chunk_size 1 succeeded
read_raw() with chunk_size 2 succeeded
read_raw() with chunk_size 3 succeeded
read_raw() with chunk_size 4 succeeded
read_raw() with chunk_size 5 succeeded
read_raw() with chunk_size 6 succeeded
read_raw() with chunk_size 7 succeeded
read_raw() with chunk_size 8 succeeded
read_raw() with chunk_size 9 succeeded
read_raw() with chunk_size 10 succeeded
read_raw() with chunk_size 11 succeeded
read_raw() with chunk_size 12 succeeded
read_raw() with chunk_size 13 succeeded
read_raw() with chunk_size 14 succeeded
read_raw() with chunk_size 15 succeeded
read_raw() with chunk_size 16 succeeded
read_raw() with chunk_size 17 succeeded
read_raw() with chunk_size 18 succeeded
read_raw() with chunk_size 19 succeeded
read_raw() with chunk_size 20 succeeded
******* provider = 'NI-VISA', data size = 2000 *******
read_raw() with chunk_size 1991 succeeded
read_raw() with chunk_size 1992 succeeded
read_raw() with chunk_size 1993 succeeded
read_raw() with chunk_size 1994 succeeded
read_raw() with chunk_size 1995 succeeded
read_raw() with chunk_size 1996 succeeded
read_raw() with chunk_size 1997 succeeded
read_raw() with chunk_size 1998 succeeded
read_raw() with chunk_size 1999 succeeded
read_raw() with chunk_size 2000 succeeded
read_raw() with chunk_size 2001 succeeded
read_raw() with chunk_size 2002 succeeded
read_raw() with chunk_size 2003 succeeded
read_raw() with chunk_size 2004 succeeded
read_raw() with chunk_size 2005 succeeded
read_raw() with chunk_size 2006 succeeded
read_raw() with chunk_size 2007 succeeded
read_raw() with chunk_size 2008 succeeded
read_raw() with chunk_size 2009 succeeded
read_raw() with chunk_size 2010 succeeded
```

MatthieuDartiailh (commented):
Thanks the change looks good. I want to audit other transports before merging in case similar issues exist there too. It may take me a couple of weeks.

berg:
I actually ran into this same bug when running some conformance tests against VPP-4.3. I have been putting LLMs to work (perhaps controversial!) trying to smoke out weird conformance/compatibility bugs between pyvisa/pyvisa-py, a Rust server implementation of HiSLIP/VXI-11 I am playing with, and some of the virtual Keysight instruments.

The same bug does exist in the HISLIP transport I believe - I was going to send along a patch for that unless this PR gets to it first.

pyvisa_py/tcpip.py
```diff
@@ -752,7 +752,10 @@ def read(self, count: int) -> Tuple[bytes, StatusCode]:
             count -= len(data)
 
             if count <= 0:
-                status = StatusCode.success_max_count_read
+                if reason & end_reason:
```
berg: When the CHR bit is set in reason (see VXI-11 RULE B.6.23), the termination char was found. I believe VPP-4.3 6.1.1 says that VI_SUCCESS_TERM_CHAR is the right return value (`StatusCode. success_termination_character_read` in pyvisa).

pyvisa_py/tcpip.py
berg, line 796: A similar bug exists here, I think, in the case that the buffer is not filled

pyvisa_py/tcpip.py
hb020: Should indeed.
There are also gaps in 
* timeout handling
* invalid `count` values
* zero length reads (which you mentioned below)

Will add that, and try to add test cases. 

I will likely not be able to handle the hislip side, as my test setup is very limited there.

pyvisa_py/tcpip.py
hb020: confirmed.

hb020:
@MatthieuDartiailh are you OK if I add more VXI-11 compliance patches in this PR? I have identified and patched the following:

* count boundary check (was absent)
* timeout handling (was incomplete)
* zero length handling (was incomplete)
* correct return value: 
   * success_max_count_read on count
   * success on (RX_END and not suppress_end_en) (the latter condition was absent)
   * success_termination_character_read on RX_CHR (was absent)

.. visible in https://github.com/hb020/pyvisa-py/tree/pr-612-extra

MatthieuDartiailh:
Tjose sounds fine to add here yes.

hb020:
merged into this PR

hb020:
DONE: Will test this more thoroughly on real HW later today. Is OK.

TODO: I might also try to set up a hislip and socket test setup. I do not have USBTMC nor VICP testbeds.

TODO: I'll add any repairs on those 2 backends (hislip/socket) here, and adapt CHANGES once done.

hb020:
@MatthieuDartiailh can we do the hislip stuff in another PR? I found a quite large load of problems in hislip, and having 3 branches open in 3 different PRs on the same source code files is not helping testing....

MatthieuDartiailh:
Yes since this PR now address multiple VXI-11 not just the original one it is fine to keep focused on this transport. I will do my best to review it in a timely manner.

hb020:
Adapted the changes list.

CHANGES
```diff
@@ -4,8 +4,14 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- Read function termination overhaul:
```
```suggestion
- VXI-11 (TCPIP::INSTR) Read function termination overhaul:
```
I will also need to overhaul this changelog before the release however since this release is huge.

pyvisa_py/tcpip.py
```diff
@@ -705,35 +714,48 @@ def read(self, count: int) -> Tuple[bytes, StatusCode]:
...
+        stop_reason = vxi11.RX_CHR
+        if not suppress_end_en:
+            stop_reason |= vxi11.RX_END
         read_fun = self.interface.device_read
         status = StatusCode.success
 
-        timeout = self._io_timeout
+        timeout = self._io_timeout  # this is derived from from self.timeout
+        # See if a timeout was set. This is derived from self.timeout, but slightly
+        # convoluted because the unit test scripts may not have set this correctly.
```
MatthieuDartiailh: Can you clarify this comment ?

hb020: OK like it is now?
```python
        # Get the timeout as cleaned up by the upper layers
        timeout = self._io_timeout

        # See if a timeout was really set. 
        # if self.timeout is None, the given timeout was VI_TMO_INFINITE
        # This lookup method is also slightly convoluted because the unit test scripts
        # may not have set this correctly.
        finite_timeout = getattr(self, "timeout", None) is not None
```

MatthieuDartiailh: I dislike the idea of making the code weirder because the existing test is poorly written. Could this be improved ?

hb020: I'll drill this down, and adapt the test scripts if needed.

hb020: should be good now.

hb020:
by the way, found no chunk size problems with hislip so far (apart from lack of events, and richness of end of message type support)

Sockets looks like it has the same problem, but also NI-Visa chokes on it.

pyvisa_py/tcpip.py
```diff
@@ -702,35 +711,51 @@ def read(self, count: int) -> Tuple[bytes, StatusCode]:
...
 
             # Calculate timeout for current chunk.
             # Limit the minimum timeout to 10ms, because 0 is undefined
             # according to VXI-11 spec (done also by Visas)
-            chunk_timeout = max(10, timeout - int((time.time() - start_time) * 1000))
+            elapsed_ms = int((time.time() - start_time) * 1000)
+            remaining_timeout = timeout - elapsed_ms
+            if finite_timeout and remaining_timeout <= 0:
+                return bytes(read_data), StatusCode.error_timeout
+            chunk_timeout = max(10, remaining_timeout)
```
MatthieuDartiailh, line 769: If there is no finite timeout this looks wrong since it will cause useless churn on the read function.

MatthieuDartiailh (commented):
I have one last concern regarding the handling of infinite timeout.

pyvisa_py/tcpip.py
hb020: understand, will look into this

hb020: pushed correction

pyvisa_py/tcpip.py
```diff
@@ -710,35 +719,54 @@ def read(self, count: int) -> Tuple[bytes, StatusCode]:
...
             # Limit the minimum timeout to 10ms, because 0 is undefined
             # according to VXI-11 spec (done also by Visas)
-            chunk_timeout = max(10, timeout - int((time.time() - start_time) * 1000))
+            if finite_timeout:
+                elapsed_ms = int((time.time() - start_time) * 1000)
+                remaining_timeout = timeout - elapsed_ms
+                if remaining_timeout <= 0:
+                    return bytes(read_data), StatusCode.error_timeout
+            else:
+                remaining_timeout = 2**32 - 1  # VI_TMO_INFINITE
```
MatthieuDartiailh: Import it and use it rather than redefining it

MatthieuDartiailh:
I wonder if the timeout handling could not be cleaned up further (not sure we need both timeout and io_timeout) but this can wait.

MatthieuDartiailh:
Thanks for this very nice work.

PR #616, opened 2026-08-11
Treat open_timeout=0 as "use the default" in TCP transports

### LINKED ISSUE #578: default open_timeout isn't used
The argument `open_timeout` for `ResourceManager.open_resource()` defaults to `0` (`constants.VI_TMO_IMMEDIATE`) if not provided:

https://github.com/pyvisa/pyvisa/blob/99d8df07d194f815ed3bb65c35903d9e96a45e01/pyvisa/highlevel.py#L3229-L3236

This causes issues with `pyvisa-py`, since it expects `None` for default values:

https://github.com/pyvisa/pyvisa-py/blob/ae0622b73c27d2100eacf08e063a197c36966bed/pyvisa_py/tcpip.py#L69-L75

This ends up causing issues with `open_timeout` for `Vxi11CoreClient`, `TCPClient`, and `RawTCPClient`:

https://github.com/pyvisa/pyvisa-py/blob/ae0622b73c27d2100eacf08e063a197c36966bed/pyvisa_py/tcpip.py#L431-L442

The default value for `open_timeout` (`5000`) isn't used since the `Vxi11CoreClient.__init__()` is called with `open_timeout=0` (or `open_timeout=None` if calling `TCPIPInstrVxi11.__init__()` directly).

This should be able to reproduce the issue with the right network conditions:

```py
import pyvisa

ip = '192.168.1.100' # change to instrument ip
resource_name = f"TCPIP::{ip}::inst0::INSTR"

try:
    # This will result in a VI_ERROR_RSRC_NFOUND error
    device = rm.open_resource(resource_name)
except pyvisa.errors.VisaIOError:
    pass

# This works since pyvisa doesn't use open_timeout before passing it to pyvisa-py
device = rm.open_resource(resource_name, open_timeout=None)
```

berg:
ResourceManager.open_resource defaults open_timeout to VI_TMO_IMMEDIATE (0), not None. But when choosing to use the default, the code checks if the timeout is not None literally. In RawTCPClient that meant _connect() got a 0 second deadline (which ends up as ~100ms) rather than the default. So any handshake slower than that failed as VI_ERROR_RSRC_NFOUND, it probably works fine generally on localhost but over the network open_timeout is effectively required.

Audit of every transport:

- VXI-11 (TCPIP::INSTR), both the portmapper and explicit-port paths, and the portmapper client itself: all funnel through RawTCPClient. Fixed.
- HiSLIP: accepted open_timeout and ignored it, hardcoding 5 s, while tcpip.py scaled it by 1000 on the way in (it is already milliseconds). Now honored on both the sync and async connections. The async socket also had no deadline at all during connect(), since it set one only afterwards.
- TCPIP::SOCKET (and the Prologix INTFC session that inherits it): was already correct; moved to the shared helper.
- VICP: pyvicp takes no separate open timeout. It derives its connect deadline from the I/O timeout, defaulting to 5 s when that is unset or 0, so the bug cannot reach it.

common.connect_timeout() is now the single place that maps a VISA open_timeout in milliseconds onto a connect deadline in seconds.

I went back and read the history of this functionality in pyvisa-py/pyvisa (may have missed some things) and I _think_ understand that the HiSLIP omission was intentional because of how VPP-4.3 is written - I read @MatthieuDartiailh's NI forum post and the spec itself, and I still (personally) think that the behavior presented in this PR is probably correct, even in light of the spec. In particular, my read of PERMISSION 4.3.2 in VPP-4.3 §4.3.3.2, which states:

> A VISA implementation MAY use the timeout parameter when opening the resource, regardless of
whether the VI_EXCLUSIVE_LOCK flag is specified.

... implies this usage is reasonable.

I'm of course very happy to drop that bit, revert back to a more strict reading of the spec (eg make the VXI-11/tcpip clients behave like hislip), or even just split the difference and just fix the None behavior.

- [x] Closes # (insert issue number if relevant)

I took that second bullet to mean that I ran the ruff/mypy checks that CI does, since it looks like the codebase migrated a while ago, and black style isnt exactly what we use anymore.

Closes #578

### DIFF
--- a/pyvisa_py/common.py
+++ b/pyvisa_py/common.py
@@ -68,13 +68,6 @@ def connect_timeout(open_timeout: Optional[float]) -> float:
         implementation should behave as if the timeout parameter is the VISA
         default timeout value of 2000 milliseconds.
 
-    That is the common path, not an edge case. ``ResourceManager.open_resource``
-    passes 0 whenever the caller supplies no ``open_timeout``. A literal 0 would
-    leave no time for the TCP handshake and fail every open with
-    ``VI_ERROR_RSRC_NFOUND``.
-
-    The cost is that ``VI_TMO_IMMEDIATE`` can no longer mean "never wait on a
-    lock". Use the session's ``lock_timeout`` for that.
     """
     return DEFAULT_OPEN_TIMEOUT / 1000 if not open_timeout else open_timeout / 1000
 
--- a/pyvisa_py/common.py
+++ b/pyvisa_py/common.py
@@ -40,16 +40,43 @@ def int_to_byte(val):
     return val.to_bytes(1, "big")
 
 
-def connect_timeout(open_timeout: Optional[float], default: float) -> float:
-    """Seconds to allow for connecting, from a VISA ``open_timeout``.
+#: Milliseconds allowed for establishing a connection when the caller does not
+#: ask for something else. See :func:`connect_timeout`.
+DEFAULT_OPEN_TIMEOUT = 2000.0
 
-    ``open_timeout`` is in milliseconds. Both ``None`` (unset) and ``0``
-    (``VI_TMO_IMMEDIATE``, which ``ResourceManager.open_resource`` passes
-    whenever the caller does not supply one) mean "use ``default``". Taking a 0
-    literally leaves no time for the TCP handshake, which turns a cold open into
-    a spurious ``VI_ERROR_RSRC_NFOUND``.
+
+def connect_timeout(open_timeout: Optional[float]) -> float:
+    """Seconds to allow for connecting, from a VISA ``open_timeout`` in ms.
+
+    ``None`` (unset) and ``0`` (``VI_TMO_IMMEDIATE``) both mean "use
+    :data:`DEFAULT_OPEN_TIMEOUT`".
+
+    VPP-4.3 section 4.3.3 defines the ``viOpen`` timeout parameter, which
+    PyVISA exposes as ``open_timeout``, as the time to wait for a lock. But
+    VPP-4.3 PERMISSION 4.3.2 also allows using it for the open itself:
+
+        A VISA implementation MAY use the timeout parameter when opening the
+        resource, regardless of whether the VI_EXCLUSIVE_LOCK flag is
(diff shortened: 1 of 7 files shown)


MatthieuDartiailh:
Thanks for your thorough investigation.
- first, yes the PR template is out of date and black is not used anymore.
- I like the idea of centralizing such conversion but I dislike the idea of confusing 0 and the default value but it aligns with the VISA specs (see below) so we can keep it.

I must have missed in the past the paragraph you quote in the VISA spec. Or I just stick to what I observed in other implementations. What bothers me is that it becomes impossible to express that one should never wait on a lock but this is not something we can fix.
Since it is permitted but not always implemented that way I would like for it to be called out explicitly in the documentation and we should migrate to the 2000 ms value mentionned in the recommendation 4.3.3.

I will do a full review ASAP but in the mean time you can explain in code (in connect_timeout the decision and link to the relevant part of the spec) and in the documentation the chosen behavior. You also need to address conflict and you can also refer to the specs in the changelog.

berg:
Thank you for taking a quick look. I'll get to work on the changes you asked for and should have something back to you tonight.

pyvisa_py/common.py
```diff
@@ -40,6 +40,45 @@ def int_to_byte(val):
...
+        implementation should behave as if the timeout parameter is the VISA
+        default timeout value of 2000 milliseconds.
+
+    That is the common path, not an edge case. ``ResourceManager.open_resource``
+    passes 0 whenever the caller supplies no ``open_timeout``. A literal 0 would
+    leave no time for the TCP handshake and fail every open with
+    ``VI_ERROR_RSRC_NFOUND``.
+
+    The cost is that ``VI_TMO_IMMEDIATE`` can no longer mean "never wait on a
+    lock". Use the session's ``lock_timeout`` for that.
```
MatthieuDartiailh:
```suggestion
```

MatthieuDartiailh (commented):
One small comment otherwise this looks good to me.

berg:
Sorry for the delay, I was away from my laptop for a few days. Accepted your change and merged main in.

PR #619, opened 2026-08-12
Allow link ID 0 to be used

### LINKED ISSUE #618: SRQ handling fails if link ID is 0
See 

```python
    def _fire_srq(self):
        try:
            # Defensive: session may have been closed while we were spawned
            if self.session.interface is None or self.session.link == 0:
                return
....
```

There is no regulation on invalid values of `Device_Link` in the VXI-11 standard. 0 is a valid value. Will create a PR.

NI-Visa, Keysight and RS all consider LID 0 to be OK.

hb020:
- [x] Closes #618 
- [x] The change is fully covered by automated unit tests. Was not tested before. Removed condition, added comment with alert, and changed test from link=1 to link=0
- [x] Documented in docs/ as appropriate. Was not documented as a limitation

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,8 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- Allow link ID 0 to be used. Closes #618 PR #619
+- VXI-11: Allow link ID 0 to be used for SRQ in agreement with the specs
+  and other implementations. Closes #618 PR #619
 - Removed return packet on device_intr_srq. Closes #614 PR #615
 - Removed SRQ server shutdown if the server was already shut down, reducing potential problems with non compliant devices. Closes #609 PR #610
 - Removed additional read_stb inside VXI-11 (TCPIP::INSTR) SRQ handling.
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- Allow link ID 0 to be used. Closes #618 PR #619
 - Removed return packet on device_intr_srq. Closes #614 PR #615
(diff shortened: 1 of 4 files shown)


MatthieuDartiailh:
Do you know if the VXI-11 specs attribute any special meaning to the link 0 ? I had no time to check yet.

hb020:
That is what I mentioned in the issue: there is no special role for 0. Any number is valid.

MatthieuDartiailh:
Sorry I went too quickly through the issue.

CHANGES
```diff
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- Allow link ID 0 to be used. Closes #618 PR #619
```
```suggestion
- VXI-11: Allow link ID 0 to be used for SRQ in agreemebt with the specs and other implementations. Closes #618 PR #619
```

pyvisa_py/tcpip.py
```diff
@@ -580,7 +580,7 @@ def close(self) -> StatusCode:
             LOGGER.error("Error closing VISA link: {}".format(e))
 
         self.interface.close()
-        self.link = 0
+        self.link = 0  # watch out: 0 is still a valid link number, do not consider 0 to be 'closed'
```
MatthieuDartiailh: I wonder if altering the link ID here make any sense since we cannot mark the link ID as invalid.

CHANGES
hb020: adapted in new commit

pyvisa_py/tcpip.py
hb020: adapted in new commit

MatthieuDartiailh:
Thanks

PR #621, opened 2026-08-13
[vxi-11] SRQ handling bug repairs: provoked instrument crash and pyvisa-py crash

### LINKED ISSUE #620: DEVICE_ENABLE_SRQ(off) provokes crash on 34465A
To Reproduce
--------------
1. Use a KS 34465A
2. provoke an SRQ (and reset the SRQ via readstb)
3. then disable the SRQ 
4. the device crashes with "Application Torreys.exe encountered a serious error and must shut down"

Latest dev of pyvisa-py.

NI-Visa does not provoke this problem.

Difference noted, maybe the cause:

NI-Visa sends a full 40 byte opaque data set in the call, while pyvisa-py sends a 0 length opaque.

A DMM6500 replies with "Parameter Error" to that call. I probably don't even have to look in the VXI-11 standard, this points to a common problem.

Will make a PR.

hb020:
**Why**

This solves 2 rather serious bugs in VXI-11 SRQ handling:

1) Disabling SRQ causes a KS 34465A to crash
2) SRQ provokes hang and exceptions in pyvisa-py when SRQ is sent by a K DMM6500

**Details**

On point 1: The body of a SRQ disable message was empty, while it appears to be expected to be always filled in with something. 

On point 2: the DMM6500 (and likely maybe others from Keithley) sends the SRQ message 
* in a stream
* with the length indicator wrong: it says it is 4 bytes longer than it really is. 
 
The code was not prepared for that, but NI-Visa has no problems with it. pyvisa-py should neither.

**Tasks**

- [x] Closes #620
- [x] The change is fully covered by automated unit tests.
- [ ] Documented in docs/ as appropriate. No doc was available on this. and these are bugs, not features.

### DIFF
--- a/pyvisa_py/protocols/vxi11.py
+++ b/pyvisa_py/protocols/vxi11.py
@@ -412,7 +412,9 @@ class SrqInterruptTCPServer(rpc.TCPServer):
         self.port = self.sock.getsockname()[1]
         stop_flag = self.session._event_state.stop_flag
         while not stop_flag.is_set():
-            LOGGER.debug("TCP SRQ (%d): Waiting for new connection from instrument", self.port)
+            LOGGER.debug(
+                "TCP SRQ (%d): Waiting for new connection from instrument", self.port
+            )
             try:
                 conn, _addr = self.sock.accept()
             except TimeoutError:
@@ -444,7 +446,9 @@ class SrqInterruptTCPServer(rpc.TCPServer):
                     # Read record marker (4 bytes)
                     marker = self._recv_all(conn, 4)
                     if marker is None:
-                        LOGGER.debug("TCP SRQ (%d): Connection closed by peer", self.port)
+                        LOGGER.debug(
+                            "TCP SRQ (%d): Connection closed by peer", self.port
+                        )
                         return  # Connection closed by peer
                 except TimeoutError:
                     continue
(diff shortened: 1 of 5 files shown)


hb020:
Found another problem with Keithley instruments: incoming events do not get detected. Timeout during read of the INTR packet. 
Will put it here, as it is roughly the same subject and I want to keep the number of open branches limited.

MatthieuDartiailh:
Fine by me

hb020:
Description of the case I added:
My Keithley DMM6500 (on the latest FW), sends out the INTR packet with a malformed content: the RPC header says it is 4 bytes longer than it in reality is. Wireshark does not detect it as a valid vxi_intr packet, but all the content is there, at the right place, and NI-Visa has no problems with it. 
So: I just accept any INTR packet that is only 4 bytes (or less) short. Now events are properly detected.

hb020:
oops, still a problem. Keithley runs the channel as a TCP stream.

hb020:
ready to go. Works on all gateways and all "serious" instruments I have.

pyvisa_py/protocols/vxi11.py
```diff
@@ -470,23 +478,68 @@ def _handle_connection(self, conn):
...
+            "TCP SRQ (%d): Receiving %d bytes from TCP SRQ connection", self.port, n
+        )
+        remaining = n
         data = b""
         while len(data) < n:
-            chunk = sock.recv(n - len(data))
+            # Some instruments omit the final 4 bytes of a VXI-11 SRQ fragment.
+            # If we already received a partial payload, treat a timeout while waiting for
+            # the last 1-4 bytes as a valid short read instead of dropping the SRQ.
+            # LOGGER.debug("TCP SRQ (%d): %d out of %d bytes remaining to receive", self.port, remaining, n)
```
MatthieuDartiailh:
```suggestion
```

MatthieuDartiailh (commented):
One leftover commented out log call.

PR #623, opened 2026-08-16
List resources return socket

### LINKED ISSUE #622: list_resources doesn't report socket devices, even when they are published via mDNS
will provide PR

hb020:
- [x] Closes #622
- [ ] The change is fully covered by automated unit tests.
- [x] Documented in docs/ as appropriate. Done

Unittest is a bit hard to do, as this requires mDNS.

This also corrects the hislip port that was hard coded to 4880, but can be something else.

I chose NOT to add vxi-11 lookups via mDNS, as portmap is more reliable.

### DIFF
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -120,8 +120,11 @@ class TCPIPInstrHiSLIP(Session):
     def list_resources(wait_time=1.0) -> List[str]:
         resources = []
         try:
-            for host in get_services("_hislip._tcp.local.", wait_time=wait_time):
-                resources.append(f"TCPIP::{host}::hislip0,4880::INSTR")
+            for host,props in get_services("_hislip._tcp.local.", wait_time=wait_time).items():
+                port = 4880
+                if "port" in props:
+                    port = props["port"]
+                resources.append(f"TCPIP::{host}::hislip0,{port}::INSTR")
         except NotImplementedError:
             warnings.warn(
                 "TCPIP::hislip resource discovery requires the zeroconf package "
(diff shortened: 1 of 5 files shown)


pyvisa_py/tcpip.py
```diff
@@ -120,8 +120,13 @@ class TCPIPInstrHiSLIP(Session):
     def list_resources(wait_time=1.0) -> List[str]:
         resources = []
         try:
-            for host in get_services("_hislip._tcp.local.", wait_time=wait_time):
-                resources.append(f"TCPIP::{host}::hislip0,4880::INSTR")
+            for host, props in get_services(
+                "_hislip._tcp.local.", wait_time=wait_time
+            ).items():
+                port = 4880
```
MatthieuDartiailh: This should be defined as a constant in the hislip (or on top of this file whichever seems best(.

pyvisa_py/tcpip.py
```diff
@@ -1257,9 +1262,24 @@ class TCPIPSocketSession(Session):
...
-        # TODO: is there a way to get this?
-        return []
+    def list_resources(wait_time=1.0) -> List[str]:
+        resources = []
+
+        try:
+            for host, props in get_services(
+                "_scpi-raw._tcp.local.", wait_time=wait_time
+            ).items():
+                port = 5025
```
MatthieuDartiailh: This should be a constant in this module (so hislip should probably be too)

MatthieuDartiailh (commented):
2 small comments. Thanks for your contribution.

pyvisa_py/tcpip.py
hb020: created default_tcpip_port local to the classes

pyvisa_py/tcpip.py
hb020: created default_tcpip_port local to the classes

pyvisa_py/tcpip.py
```diff
@@ -1254,10 +1261,27 @@ class TCPIPSocketSession(Session):
     # for a specific kind of resource
     parsed: rname.TCPIPSocket
 
+    default_tcpip_port = 5025
+
     @staticmethod
-    def list_resources() -> List[str]:
-        # TODO: is there a way to get this?
-        return []
+    def list_resources(wait_time=1.0) -> List[str]:
```
MatthieuDartiailh, line 1267: We should double check if the VISA query can allow to lock up resources on different port than the default one and if yes plan to support it.

MatthieuDartiailh (approved):

pyvisa_py/tcpip.py
hb020: VXI11: the port is dynamic. Even if it is published on a fixed port via mDNS. That is why I propose to keep vxi11 discovery via portmap.

Sockets: always the same port.

MatthieuDartiailh: The port used for socket resources is part of the resource name and I am pretty sure I have seen ports other than 5025 used by some manufacturers.

hb020: > The port used for socket resources is part of the resource name and I am pretty sure I have seen ports other than 5025 used by some manufacturers.

yes, definitely. I meant: the port will not change dynamically.

PR #631, opened 2026-08-25
[vxi-11] Give the END indicator priority over the termination character

berg:
I missed getting a comment in about this before #612 was merged but I don't think the current behavior is correct:

A VXI-11 read that stops on both the END indicator and the termination character reported VI_SUCCESS_TERM_CHAR. VPP-4.3 RULE 6.1.1 requires VI_SUCCESS whenever END is received and VI_ATTR_SUPPRESS_END_EN is clear, regardless of whether the termination character was also read.

Instruments commonly assert END on the byte that carries the termination character, so the two arrive together in one reason and the RX_CHR test, which came first, always won.

VI_SUCCESS_TERM_CHAR is also no longer reported while VI_ATTR_TERMCHAR_EN is clear, which RULE 6.1.5 forbids. OP_FLAG_TERMCHAR_SET is only sent when the attribute is set, so this guards against an instrument that reports RX_CHR anyway.

cc @hb020

### DIFF
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -693,12 +693,20 @@ class TCPIPInstrVxi11(Session):
                 pass
 
     def _read_status_from_reason(
-        self, reason: int, suppress_end_en: bool
+        self, reason: int, suppress_end_en: bool, termchar_en: bool
     ) -> StatusCode:
-        if reason & vxi11.RX_CHR:
-            return StatusCode.success_termination_character_read
+        """Completion status for a read, from the reason the device gave.
+
+        VPP-4.3 RULE 6.1.1 gives the END indicator priority: it is answered
+        regardless of whether the termination character was also read, so it
+        is tested first. RULE 6.1.4 withholds VI_SUCCESS while
+        VI_ATTR_SUPPRESS_END_EN is set, and RULE 6.1.5 withholds
+        VI_SUCCESS_TERM_CHAR while VI_ATTR_TERMCHAR_EN is clear.
+        """
         if reason & vxi11.RX_END and not suppress_end_en:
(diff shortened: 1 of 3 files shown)


hb020:
looks good indeed, sorry I missed that.

pyvisa_py/tcpip.py
```diff
@@ -693,12 +693,20 @@ def _stop_event_monitor(self) -> None:
...
-            return StatusCode.success_termination_character_read
+        """Completion status for a read, from the reason the device gave.
+
+        VPP-4.3 RULE 6.1.1 gives the END indicator priority: it is answered
+        regardless of whether the termination character was also read, so it
+        is tested first. RULE 6.1.4 withholds VI_SUCCESS while
+        VI_ATTR_SUPPRESS_END_EN is set, and RULE 6.1.5 withholds
+        VI_SUCCESS_TERM_CHAR while VI_ATTR_TERMCHAR_EN is clear.
+        """
         if reason & vxi11.RX_END and not suppress_end_en:
```
MatthieuDartiailh:
```suggestion
        if not suppress_end_en and (reason & vxi11.RX_END):
```

pyvisa_py/tcpip.py
```diff
@@ -693,12 +693,20 @@ def _stop_event_monitor(self) -> None:
...
             return StatusCode.success
+        if reason & vxi11.RX_CHR and termchar_en:
```
MatthieuDartiailh:
```suggestion
        if termchar_en and (reason & vxi11.RX_CHR):
```

MatthieuDartiailh (commented):
2 formatting nitpick (avoid doing work when option is false) otherwise LGTM

PR #632, opened 2026-08-25
Set the TCP keepalive probe timers only where they exist

berg:
VI_ATTR_TCPIP_KEEPALIVE set TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT without checking that the running interpreter has them. TCP_KEEPIDLE does not exist on macOS, which calls the same option TCP_KEEPALIVE, so enabling keepalive there raised AttributeError for VXI-11 and TCPIP::SOCKET resources.

The three call sites now share common.set_keepalive, which applies each probe timer only if it is available and always sets SO_KEEPALIVE, so the attribute does what it says on every platform and tunes the timers where it can.

- [x] Documented in docs/ as appropriate -> I don't think any docs are needed here 

### DIFF
--- a/pyvisa_py/common.py
+++ b/pyvisa_py/common.py
@@ -7,6 +7,7 @@
 """
 
 import logging
+import socket
 from typing import TYPE_CHECKING, Iterator, Optional
 
 from pyvisa import logger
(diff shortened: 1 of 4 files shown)


berg:
This lets mypy correctly run on macOS (which also allows changes to tcpip.py to have pre-commit hooks run properly there).

pyvisa_py/common.py
```diff
@@ -21,6 +22,34 @@
...
+        return
+
+    for name, value in (
+        ("TCP_KEEPIDLE", 60),
+        ("TCP_KEEPALIVE", 60),  # macOS spelling of TCP_KEEPIDLE
+        ("TCP_KEEPINTVL", 60),
+        ("TCP_KEEPCNT", 5),
+    ):
+        option = getattr(socket, name, None)
+        if option is None:
```
MatthieuDartiailh, line 45:
```suggestion
        if not hasattr(socket, name):
```

berg: We use `option` below, so I used `getattr` here to avoid calling it twice (hasattr is generally just `getattr` wrapped in try/except, I think?)

Happy to change this but I think this is more readable than changing the line below to

```
if not hasattr(socket, name):
    sock.setsockopt(socket.IPPROTO_TCP, getattr(socket, name), value)

# or

if not hasattr(socket, name):
    option = getattr(socket, name)
    sock.setsockopt(socket.IPPROTO_TCP, option, value)
```

MatthieuDartiailh: My bad sorry for the noise

PR #626, opened 2026-08-17
Events on hislip

hb020:
- [x] Closes #625
- [x] Executed ``black . && isort -c . && flake8`` with no errors. 
- [x] The change is fully covered by automated unit tests
- [ ] Documented in docs/ as appropriate. -> preparing that for another PR.
- [x] Added an entry to the CHANGES file

PR #633, opened 2026-08-25
VXI11 lock handling

### LINKED ISSUE #583: VXI-11 lock handling incorrect
The VXI-11 lock handling is in general incorrect.

The SocketTimeout is set to a smaller value as the actual requested lock_timeout.
This results into a break of the communication, especially when exclusive locks are used.

The socket timeout has to be set in alignment with lock_timeout _(if the called function has it)_+ io_timeout _(if the called function has it)_ + "some" extra margin.

To Reproduce
--------------

```
VISANAME = 'TCPIP::192.168.3.2::inst22::INSTR'

import pyvisa as visa
import time
rm = visa.ResourceManager('@py')
dev = rm.open_resource(VISANAME)
dev.lock()

dev2 = rm.open_resource(VISANAME) # expectation: this fails with error 11 after the 10000ms of lock time, however it fails after ~7s with sockettimeout
```

Wireshark trace:

Image

1. Here you see that pyvisa-py requested 10s lock timeout.
2. You see that between packet 224 and 227 10s pass. However, pyvisa-py raises an exception "Exception: error creating link: 3" BEFORE the actual 10s lock timeout passed and the reply was sent.

This is just one occurance of that issue. Note: create_link has no flags field, that means there is no waitforlock information. That means the lock timeout of 10s pyvisa-py passes always has to be respected when the device is locked.

I have no immediate bandwidth to provide a PR as this can be a tricky one, but might be able to pick it up at a later stage.

hb020:
**Why**

This allows more compliant lock handling on VXI-11.

- [x] Closes #583

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -233,7 +233,7 @@ in case another program or session has locked it, you must choose one of the fol
     that already has a lock: they should reply immediately with ``VI_ERROR_RSRC_LOCKED``.
     However, in practice this is not always the case, and some instruments take quite some liberties with it.    
 
-    ``VI_KTATTR_LOCKWAIT`` may not be available yet in PyVISA. In that case, you could do this:
+    ``pyvisa.constants.VI_KTATTR_LOCKWAIT`` may not be available yet in PyVISA. In that case, you could do this:
 
     >>> import pyvisa
     >>> rm = pyvisa.ResourceManager('@py')
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -176,7 +176,7 @@ Locking
 **PyVISA-Py only supports exclusive locking, and only on VXI-11. Shared locks and nested locking are not supported.** 
 
 Socket instruments (``TCPIP::SOCKET``) do not support locking.
-HiSLIP instruments (``TCPIP::hislip0``) could support locking, but PyVISA-Py does not yet implement this feature.
+HiSLIP instruments (``TCPIP::hislip``) could support locking, but PyVISA-Py does not yet implement this feature.
 
 With exclusive locking, only one session can be used at a time on an instrument.  
 If another session has a lock, another client will not be able to communicate with
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -63,13 +63,13 @@ Are GBIP secondary addresses supported?
 GPIB secondary addresses are supported in NI-VISA fashion, meaning that the
 secondary address is not 96 to 126 as transmitted on the bus, but 0 to 30.
 
-For expample, `GPIB0::9::1::INSTR` is the address of the first VXI module
-controlled by a GPIB VXI command module set to primary address `9`, while
-the command module itself is found at `GPIB0::9::0::INSTR`, which is distinct
-from a pure primary address like `GPIB0::9::INSTR`.
+For expample, ``GPIB0::9::1::INSTR`` is the address of the first VXI module
+controlled by a GPIB VXI command module set to primary address ``9``, while
+the command module itself is found at ``GPIB0::9::0::INSTR``, which is distinct
+from a pure primary address like ``GPIB0::9::INSTR``.
 
 ``ResourceManager.list_resources()`` can discover both primary and secondary 
-addressable `GPIB::...::INSTR` resources. As a result, it can be slow,
+addressable ``GPIB::...::INSTR`` resources. As a result, it can be slow,
 as it now needs to check 992 addresses per GPIB controller instead of just 31.
 
 For every primary address where no listener is detected, all
@@ -82,12 +82,12 @@ checked as most devices simply ignore secondary addressing.
 If you have an instrument that reacts to the primary address and has different
 functionality on some secondary addresses, please leave a bug report.
 
-If you use a VXI-11.2 gateway (VXI-11 to GPIB), you can use constructions like 
-`TCPIP::host::gpib0,9,1::INSTR`, where `gpib0` is the 'GPIB SICL Interface Name' 
-configured on the gateway, `9` is the primary address of the instrument, 
-and `1` is the secondary address of that instrument.
+If you use a VXI-11.2 (VXI-11 to GPIB) gateway, you can use constructions like 
+``TCPIP::host::gpib0,9,1::INSTR``, where ``gpib0`` is the 'GPIB SICL Interface Name' 
+configured on the gateway, ``9`` is the primary address of the instrument, 
+and ``1`` is the secondary address of that instrument.
 ``ResourceManager.list_resources()`` will however not try to discover any resources
-on a VXI-11.2 gateway, as the SICL Interface Name is not automatically known.
+behind a VXI-11.2 gateway, as the SICL Interface Name is not automatically known.
 
 
 Can PyVISA-py be used from a VM?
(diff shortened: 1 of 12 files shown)


hb020:
releasing this for merge. hislip is more effort, will come in a next PR.

docs/source/faq.rst
```diff
@@ -177,9 +169,92 @@ omit the argument.
...
+    >>> # allow 3 s to reach the instrument
+    >>> inst = rm.open_resource('TCPIP::192.168.1.100::INSTR', open_timeout=3000)
+    >>> # and then try to acquire a lock on the instrument with a 10 s timeout
+    >>> inst.lock_excl(timeout=10000)
+
+If you have not locked the instrument, and want to control the behaviour of your program
+in case another program or session has locked it, you must choose one of the following methods:
+
+- Request a lock via ``inst.lock_excl()``. This is the most portable. See above.
+- Configure the lock timeout via the PyVISA-Py specific ``rm.visalib.sessions[inst.session].lock_timeout``.
```
MatthieuDartiailh: Why do you need to poke at the internals here ? I would prefer to not encourage such uses.
Since Keysight make use of a non standard attribute to control this parameter, we could do the same. This ways internals are not exposed.

pyvisa_py/tcpip.py
```diff
@@ -829,21 +889,25 @@ def write(self, data: bytes) -> Tuple[int, StatusCode]:
             num = len(data)
             offset = 0
 
+            # TODO determine how OP_FLAG_WAIT_BLOCK could be set from the outside.
+            # VPP-4.3 does not provide a clear definition for it.
+            # For now, we derive it from self.lock_timeout.
+            lock_timeout = self.lock_timeout
+            if lock_timeout != constants.VI_TMO_IMMEDIATE:
+                flags |= vxi11.OP_FLAG_WAIT_BLOCK
+
```
MatthieuDartiailh: This could be factored out in a helper.

MatthieuDartiailh (commented):
First quick pass.

pyvisa_py/tcpip.py
hb020: yes, will do

docs/source/faq.rst
hb020: well, it was like that, so I left it. But yes, I prefer the keysight way.

BUT: then we also need to adapt pyvisa, since it is pyvisa that would expose that specific attribute to the user. It is not a standard attribute.

MatthieuDartiailh: I do not have all the code within easy reach to check but since pyvisa mostly forward directly to the library for get/set attribute we may be able to to define the extra args outside PyVISA.

hb020: sure, can be done, the other attributes are 'hard coded' in `pyvisa/constants.py`. They do not refer to the backends.
Not a big issue, but then we'll have to make sure to catch any mismatching package versions.

pyvisa_py/tcpip.py
hb020: done

docs/source/faq.rst
hb020: I'll work on something. For now, only in pyvisa-py. And making sure it can live with outdated pyvisa.

MatthieuDartiailh (commented):
I would expect changes to be required for serial and USB for locking when opening. And for those and GPIB since locking does not do anything we should reject non default value.

hb020:
Implemented VI_KTATTR_LOCKWAIT, removed lock_timeout. It is all limited it to vxi-11

hb020:
In fact, VPP-4.3 states that all types should support locking, both exclusive locks and shared locks, AND nested locking. 

For exclusive locks, VXI-11 and HiSLIP instruments handle that themselves (maybe also vicp and vxi?). 
The other instrument interfaces do not support it AFAIK.
Also, shared locking, and nested locking, mean a fair amount of work inside the higher layers of pyvisa-py. Much more than this. Seen potential low return (those who use shared locking, often use large setups, and they use NI-VISA), not sure I will get to that one day.

pyvisa_py/tcpip.py
```diff
@@ -931,24 +1006,30 @@ def assert_trigger(self, protocol: constants.TriggerProtocol):
             Return value of the library call.
 
         """
+        flags = 0
+        flags, lock_timeout = self._adapt_flags_and_lock_timeout(flags)
+
         # XXX make this nicer (either validate protocol or pass it)
```
MatthieuDartiailh: Not sure this comment still applies

hb020:
Now I need to work on the timeout of the remote/local functions

MatthieuDartiailh:
> I would expect changes to be required for serial and USB for locking when opening. And for those and GPIB since locking does not do anything we should reject non default value.

I realize this was silly since serial is by definition a unique connection and USBTMC and GPIB are likely the same. I will make another more thorough review but I think are nearly good.

hb020:
merged.
Preferred order of integration: this PR and only then #626. I might need to do some merging and cleaning for that one, and it is less important.

pyvisa_py/tcpip.py
```diff
@@ -509,7 +525,14 @@ class TCPIPInstrVxi11(Session):
...
     #: Time to wait in ms before erroring with a timeout when trying to acquire a lock
-    lock_timeout: int = 10000
+    # 0 = immediate, >0 = wait for that many milliseconds
+    # This is used for most operations (mainly except open_resource and lock_excl)
+    #
+    # NI-VISA and R&S VISA do not expose any of this.
+    # Keysight exposes the Keysight-specific VISA ViBoolean local (per-session) attribute VI_KTATTR_LOCKWAIT
+    #   When False, when already locked, immediately returns VI_ERROR_RSRC_LOCKED
+    #   When True, uses lock timeout = session timeout interval. When locked and timed out, returns VI_ERROR_TMO
+    # lock_timeout: int = 0
```
MatthieuDartiailh: This does not really belong here anymore.

docs/source/faq.rst
```diff
@@ -177,9 +169,112 @@ omit the argument.
...
+    >>> # Define the raw hex constant for VI_KTATTR_LOCKWAIT
+    >>> VI_KTATTR_LOCKWAIT = 0x0FFF002BL
+    >>>
+    >>> # Set lockwait to True (VI_TRUE = 1)
+    >>> inst.set_attribute(VI_KTATTR_LOCKWAIT, 1)
+    >>> # Read back the attribute value
+    >>> lockwait_val = inst.get_attribute(VI_KTATTR_LOCKWAIT)
+    >>> print("Lockwait state:", lockwait_val)
+    >>>
+    >>> # Do your operations
```
MatthieuDartiailh: Why do you need this since we force set this attribute?

pyvisa_py/tcpip.py
MatthieuDartiailh: This whole paragraph seems out of place.

MatthieuDartiailh (commented):
Some small comments

docs/source/faq.rst
hb020: Otherwise you get this: ``NameError: name 'VI_KTATTR_LOCKWAIT' is not defined``

It needs to be added to pyvisa.constants

pyvisa_py/tcpip.py
hb020: yes, will remove the comments

docs/source/faq.rst
hb020: clarified faq

pyvisa_py/tcpip.py
hb020: done

pyvisa_py/tcpip.py
hb020: corrected comment

docs/source/faq.rst
hb020: oops. you're right. WIll correct the faq.

hb020: corrected

PR #637, opened 2026-08-26
Check SO_ERROR before reporting a connection established

berg:
_connect drives a non-blocking connect and waits for the result with select. It assumed that once the socket was ready it was connected. But a refused connection also makes the socket ready, with the reason in SO_ERROR, and the connection would just fail on the first send. viOpen raised an OSError (BrokenPipeError on macOS, ConnectionRefusedError on Linux), rather than a VisaIOError, which complicates error handling.

```python
import pyvisa

# Nothing is listening on port 9999. The ",9999" names the VXI-11 core channel
# port directly, so this does not depend on a portmapper being reachable.
try:
    pyvisa.ResourceManager("@py").open_resource(
        "TCPIP::127.0.0.1,9999::inst0::INSTR", open_timeout=2000
    )
except pyvisa.errors.VisaIOError as exc:
    print("VisaIOError:", exc.abbreviation)
except Exception as exc:
    print("not a VISA error:", type(exc).__name__, exc)
```

Before, on macOS 24.6.0 / Python 3.14.7:

```
not a VISA error: BrokenPipeError [Errno 32] Broken pipe
```

and on Linux 6.12.13 / Python 3.13.15:

```
not a VISA error: ConnectionRefusedError [Errno 111] Connection refused
```

After, on both:

```
VisaIOError: VI_ERROR_RSRC_NFOUND
```

The original errno from the socket failure is sent to the debug logger.

- [ ] Documented in docs/ as appropriate -> does this need documentation?

### DIFF
--- a/pyvisa_py/testsuite/test_rpc.py
+++ b/pyvisa_py/testsuite/test_rpc.py
@@ -1,0 +1,62 @@
+"""Tests for the ONC RPC layer used by VXI-11.
+
+:copyright: 2014-2024 by PyVISA-py Authors, see AUTHORS for more details.
+:license: MIT, see LICENSE for more details.
+
+"""
+
+import socket
+
+import pytest
+
+from pyvisa_py.protocols import rpc
+
+
+@pytest.fixture
+def closed_port():
+    """A port on the loopback interface with nothing listening on it."""
+    probe = socket.socket()
+    probe.bind(("127.0.0.1", 0))
+    port = probe.getsockname()[1]
+    probe.close()
+    return port
+
+
+@pytest.fixture
+def listening_port():
+    """A port that accepts connections. Yields (port, server socket)."""
+    server = socket.socket()
+    server.bind(("127.0.0.1", 0))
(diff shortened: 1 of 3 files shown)


berg:
as a meta note, I have a bunch of smallish papercut fixes like this, but I've been trickling them in because they all end up conflicting around the CHANGES file - if it's easier for me to submit them as (a) a PR with multiple commits, at least grouped around a theme or (b) a few all at once even though I'll need to update them as they are merged, let me know!

MatthieuDartiailh:
I am fine with slightly larger PR. You could also give stashed PR a try since it seems like an appropriate use case.

pyvisa_py/testsuite/test_rpc.py
MatthieuDartiailh, line 1: I will have to consolidate those tests because we have now many files but working on testing is the next item on my to-do list for PyVISA projects anyway.

MatthieuDartiailh (approved):

Issue #638, opened 2026-08-29
VXI-11, HiSLIP,..  support wrap up: loose ends


hb020:
Still TODO:

* VXI-11: 
  * add viTerminate/abort
* investigate gaps remaining:
  * events (check VPP-4.3)
  * locking on serial/socket (check VPP-4.3)
  * mandatory attributes (VPP-4.3 page 5-66 .. 5-73)
* document gaps remaining:
  * VXI-11+HiSLIP: 
    * shared locking and nested locking, 
    * authentication
  * secure channel on HiSLIP
* document additions: 
  * VXI-11 docmd
  * prologix with howto (see https://github.com/pyvisa/pyvisa/issues/112)
  * trigger on prologix

doCMD example, _first check what backends support it_:
```python
import pyvisa

# 1. Force the pure-Python backend (@py)
rm = pyvisa.ResourceManager('@py')

# 2. Open your instrument session
# (Replace with your actual GPIB, USB, or TCPIP resource string)
inst = rm.open_resource('TCPIP::192.168.1.20::gpib0,1::INSTR')

# 3. Access device_docmd via the underlying visalib
# Syntax: inst.visalib.device_docmd(session_handle, cmd, value, fmt)
session_handle = inst.session
cmd = 0x08        # Example IEEE 488 command integer
value = True      # Associated payload value
fmt = '?'         # Struct-style payload format character (e.g., '?' for bool, 'i' for int)

response = inst.visalib.device_docmd(session_handle, cmd, value, fmt)
print(response)
```

hb020:
@MatthieuDartiailh 

The following compliance report was generated by copilot. I'm on the fence about this:

* to be validated manually and to be put in context. Some of the findings are wrong.
* a lot of text, maintenance might not be easy
* are we sure we want to show all that "missing"? Probably not. Just mention what is supported and partially supported is the most important I think.

I'll continue on this, will add the major functions and their limitations, if any.

And will add the additions (prologix, the keysight attribute,..)

[compliance.html](https://github.com/user-attachments/files/31696916/compliance.html)

hb020:
Working on documentation in https://github.com/hb020/pyvisa-py/blob/unknown_attribute/docs/source/compliance.rst

berg:
hey @hb020 - if it is helpful, i ported the internal thing i'd been using over to using github actions:
https://berg.github.io/testgear-network-stress/

[...]

PR #639, opened 2026-08-29
improve triggers

### LINKED ISSUE #634: [multiple instrument types] viAssertTrigger ignores the trigger protocol argument
This is what copilot comes with:

> **viAssertTrigger** ignores the trigger protocol argument
> assert_trigger() accepts a protocol parameter but ignores it in tcpip.py. It unconditionally calls the device-level trigger operation.
> 
> Under VPP-4.3, a trigger protocol argument must be validated. Unsupported protocols should be rejected rather than silently mapped to the default trigger path. The implementation does not enforce that at all.
> 
> So this is a layer mismatch between VISA API semantics and the underlying VXI-11 call.

Not validated yet, I will look into this. Just noting it here as a TODO.

hb020:
Combination:
* Triggers
* effects of VI_ATTR_IO_PROT (as there are some outside of triggers)

And I removed the HiSLIP casts.

- [x] Closes #634
- [x] The change is fully covered by automated unit tests. 

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,9 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- VXI-11 and HiSLIP: added filtering on trigger, added trigger to HiSLIP, and doc #634 PR #639
+- VXI-11 and HiSLIP: added filtering on trigger,
+  added trigger to HiSLIP, serial and socket,
+  and documented. #634 PR #639
 - VXI-11 (TCPIP::INSTR):
   Locking compliance improvement. Exclusive locking works now,
   Removed `session.lock_timeout`, no longer needed for instruments
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- VXI-11 and HiSLIP: added filtering on trigger, added trigger to HiSLIP, and doc #634 PR #6??
+- VXI-11 and HiSLIP: added filtering on trigger, added trigger to HiSLIP, and doc #634 PR #639
 - VXI-11 (TCPIP::INSTR):
   Locking compliance improvement. Exclusive locking works now,
   Removed `session.lock_timeout`, no longer needed for instruments
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,7 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- VXI-11 and HiSLIP: added filtering on trigger, added trigger to HiSLIP, and doc #634 PR #6??
 - VXI-11 (TCPIP::INSTR):
   Locking compliance improvement. Exclusive locking works now,
   Removed `session.lock_timeout`, no longer needed for instruments
--- a/CHANGES
+++ b/CHANGES
@@ -4,12 +4,14 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- VI_ATTR_IO_PROT, and triggering:
-  VXI-11 and HiSLIP: added filtering on trigger,
-  added trigger to HiSLIP, serial and socket,
+- Triggers:
+  Added trigger to HiSLIP, serial and socket
+  Added filtering to VXI-11 triggers
+  Closes #634 PR #639
+- VI_ATTR_IO_PROT:
   added or completed clear to serial and socket,
   added readstb to serial and socket,
-  and documented. #634 PR #639
+  and documented. Closes #634 PR #639
 - VXI-11 (TCPIP::INSTR):
   Locking compliance improvement. Exclusive locking works now,
   Removed `session.lock_timeout`, no longer needed for instruments
--- a/CHANGES
+++ b/CHANGES
@@ -4,8 +4,11 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
-- VXI-11 and HiSLIP: added filtering on trigger,
+- VI_ATTR_IO_PROT, and triggering:
+  VXI-11 and HiSLIP: added filtering on trigger,
   added trigger to HiSLIP, serial and socket,
+  added or completed clear to serial and socket,
+  added readstb to serial and socket,
   and documented. #634 PR #639
 - VXI-11 (TCPIP::INSTR):
   Locking compliance improvement. Exclusive locking works now,
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -319,7 +319,7 @@ In addition, for Prologix resources: ``inst.assert_trigger()`` will send ``++trg
 Triggers are however not supported for Serial devices or TCP/IP sockets. 
 Just send the string ``*TRG\n`` if needed. 
 (Know that they would only be able to accept a trigger if you set ``VI_ATTR_IO_PROT``
-to the right value, and even then, that would just mean that ``*TRG\n`` is to be sent.)
+to ``VI_PROT_4882_STRS``, and even then, that would just mean that ``*TRG\n`` is to be sent.)
 
 
 
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -317,7 +317,10 @@ USB and USBTMC resources that support it.
 In addition, for Prologix resources: ``inst.assert_trigger()`` will send ``++trg\n``
 
 Triggers are however not supported for Serial devices or TCP/IP sockets. 
-Just send the string ``*TRG\n`` if needed.
+Just send the string ``*TRG\n`` if needed. 
+(Know that they would only be able to accept a trigger if you set ``VI_ATTR_IO_PROT``
+to the right value, and even then, that would just mean that ``*TRG\n`` is to be sent.)
+
 
 
 .. _PySerial: https://pythonhosted.org/pyserial/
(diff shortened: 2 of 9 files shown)


MatthieuDartiailh (commented):
You removed the cast but did not change the type of interface. What did I miss?

hb020:
well, the interface type was OK already: line 137

And some used cast, other didn't. Couldn't find a reason.

docs/source/faq.rst
```diff
@@ -307,6 +307,18 @@ and what NI-VISA does, so this should be fully portable.
...
+--------
+
+The trigger functionality in PyVISA-Py is almost feature complete. 
+Triggers are supported on GPIB, VXI-11 and HiSLIP resources, and on the 
+USB and USBTMC resources that support it.
+
+In addition, for Prologix resources: ``inst.assert_trigger()`` will send ``++trg\n``
+
+Triggers are however not supported for Serial devices or TCP/IP sockets. 
+Just send the string ``*TRG\n`` if needed.
```
MatthieuDartiailh: Hardware trigger are not part of VISA for raw socket. For serial I am not sure.

hb020: hardware trigger CAN be supported for serial and sockets:

> For a session to a Serial device or TCP/IP socket, if VI_ATTR_IO_PROT is VI_PROT_4882_STRS, the device is sent the string “*TRG\n”; otherwise, this operation is not valid.

Very convoluted. It would mean adding support for that attribute, and documentation for something that can be easily bypassed, and will likely never be used.

Expanded the FAQ with that info.

MatthieuDartiailh: Pyvisa does support this attribute so adding support on PyVISA-py side should not be too hard and would make sense to be standard compliant.

hb020: OK, will go for it. PR on hold until then

hb020:
BTW, we should move python upper limit to 3.14 and 3.15-dev (and in one month to 3.15).

Also, tried including pyserial via the ci file to allow fuller pytest, and then windows checks failed in `test_serial.py`. Don't have windows here, so can't debug that.

docs/source/faq.rst
```diff
@@ -307,6 +307,19 @@ and what NI-VISA does, so this should be fully portable.
 GPIB has functionality comparable to HiSLIP, but may behave differently than NI-VISA, 
 depending on the type of interface.
 
+Triggers
+--------
+
+The trigger functionality in PyVISA-Py is almost feature complete.
```
MatthieuDartiailh, line 313: Feel like we have done everything possible no ?

pyvisa_py/tcpip.py
```diff
@@ -377,6 +376,28 @@ def clear(self) -> StatusCode:
...
+        Returns
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        if protocol != constants.TriggerProtocol.default:
+            return StatusCode.error_nonsupported_operation
+
+        self.interface.trigger()
```
MatthieuDartiailh, line 400: Shouldn't we inspect the io_prot attribute? If the standard sys no we should quote it.

MatthieuDartiailh (commented):
I wonder if we should not always inspect io_prot and respect its value.

docs/source/faq.rst
hb020: not sure about vicp, gpib and usbtmc. The first 2 I don;t have, and the last one I only have one beta version device under construction, so level of confidence low. Just want to be cautious.

pyvisa_py/tcpip.py
hb020: After a look in the standard, you are right. 

> Specifies which protocol to use. In VXI systems, for
example, you can choose between normal word serial or fast
data channel (FDC). In GPIB, you can choose between
normal and high-speed (HS488) data transfers. In ASRL and
TCPIP systems, you can choose between normal and 488-
style transfers, in which case the viAssertTrigger() and
viReadSTB() operations send 488.2-defined strings.

I could add that in this PR, but then we'll touch much more code. Subject for a separate PR?

MatthieuDartiailh: I think you can include in this PR. 
Do you know what would normal mode be for serial ?

hb020: normal (as for socket) = no support of "out of band" messages, mainly trigger, stb, srq and terminate. In the case of `io_prot`, transforming them to in-band means translating to IEEE-488.2 in-band equivalents for trigger and stb: `*TRG` and `*STB?`. SRQ cannot be done since it is in the other direction, and terminate is by nature impossible in-band.

In other words: 'normal' means no trigger and no readstb.

hb020: tracked all io_prot use that I can see in VPP-4.3. Have no GPIB HW, so I cannot check the effect on GPIB resources.

CHANGES
```diff
@@ -4,6 +4,12 @@ PyVISA-py Changelog
 0.9.0 (unreleased)
 ------------------
 
+- VI_ATTR_IO_PROT, and triggering:
+  VXI-11 and HiSLIP: added filtering on trigger,
+  added trigger to HiSLIP, serial and socket,
+  added or completed clear to serial and socket,
+  added readstb to serial and socket,
```
MatthieuDartiailh, line 13: This should be a separate entry.

docs/source/faq.rst
```diff
@@ -307,6 +307,36 @@ and what NI-VISA does, so this should be fully portable.
...
+
+- Triggers will be sent using the ``*TRG\n`` command.
+- Clear will be sent using the ``*CLS\n`` command.
+- Read STB will be done using the ``*STB?\n`` query command.
+
+Any other value will not be recognized and may result in an error.
+
+This is conform to VPP-4.3 and should therefore be portable.
+
+For GPIB resources, this attribute is not (yet) supported.
```
MatthieuDartiailh, line 329: This bothers me a bit since it duplicates the official VISA specs.

pyvisa_py/gpib.py
```diff
@@ -886,6 +889,9 @@ def _set_attribute(
                 return StatusCode.success
             except gpib.GpibError:
                 return StatusCode.error_nonsupported_attribute_state
+        elif attribute == constants.VI_ATTR_IO_PROT:
```
MatthieuDartiailh, line 892: Could we support setting the attribute to normal ?

pyvisa_py/serial.py
```diff
@@ -197,6 +198,26 @@ def write(self, data: bytes) -> Tuple[int, StatusCode]:
...
+        # VPP 4.3:
+        # For Serial INSTR sessions, VISA must flush (discard) the I/O output buffer, send a
+        # break, and then flush (discard) the I/O input buffer.
+        self.interface.reset_output_buffer()
+        self.interface.sendBreak()
+        self.interface.reset_input_buffer()
+
+        # Send the *CLS command to clear the device if using the 488.2 STRS protocol
+        # VPP-4.3 Permission 6.14
+        if self.attrs[ResourceAttribute.io_prot] == constants.VI_PROT_4882_STRS:
```
MatthieuDartiailh, line 216: Shouldn't we error if the value is something else like we do for other magic methods ?

pyvisa_py/tcpip.py
```diff
@@ -377,6 +376,28 @@ def clear(self) -> StatusCode:
...
+        protocol : constants.TriggerProtocol
+            Trigger protocol to use during assertion. Only default is supported.
+
+        Returns
+        -------
+        StatusCode
+            Return value of the library call.
+
+        """
+        if protocol != constants.TriggerProtocol.default:
```
MatthieuDartiailh, line 395: I would expect to see support for IEEE 488.2 standard string too.

docs/source/faq.rst
hb020: indeed, was looking how to put this (see #638 ). Am planning to create an overall compliance table:
functionality and attributes vs resource type.

Then the effects of  VI_PROT_4882_STRS will no longer be shown, only the fact that it is supported.

CHANGES
hb020: ok, will adapt

pyvisa_py/gpib.py
hb020: yes, will add that.

pyvisa_py/tcpip.py
hb020: that is not explicitly mentioned in the standard, so I left that out.

pyvisa_py/serial.py
hb020: The standard does not say that you shouldn't allow setting it to an unused value. Just that you should complain the moment you want to use it and it has an invalid value.

Also: it explicitly says:

> OBSERVATION 6.1.23
> The viClear() operation will no longer return an error for a Serial INSTR resource or a TCP/IP
> SOCKET resource when the attribute VI_ATTR_IO_PROT is set to VI_PROT_NORMAL.

I can add that remark in the code, to clarify.

MatthieuDartiailh: Yes please

pyvisa_py/tcpip.py
```diff
@@ -1074,7 +1095,10 @@ def assert_trigger(self, protocol: constants.TriggerProtocol):
             Return value of the library call.
 
         """
-        # protocol is ignored, VXI-11 doesn't support multiple types
+        if protocol != constants.TriggerProtocol.default:
+            return StatusCode.error_nonsupported_operation
+
+        # Contrary to serial and socket, io_prot is not used, because VPP-4.3 does not explicitly demands it.
```
MatthieuDartiailh, line 1101: I would be curious to see what other impl do here.

MatthieuDartiailh (approved):
One thing we could check what other do but otherwise LGTM.

pyvisa_py/tcpip.py
hb020: NI-VISA does like PyVISA-Py here.

MatthieuDartiailh: Thanks for checking

PR #643, opened 2026-09-01
Lock on hislip

### LINKED ISSUE #635: Allow locking on hislip
This should include 
* support of lock-on-open (accessmode), by chaining AsyncLock after the init sequence
* lock after open (AsyncLock)
* test suite
* adaptations in FAQ

and (note to self) do not expect writes to be blocked if someone else has a lock, as it it will be scheduled for later by the instrument. Reading a reply will however just hang.

hb020:
Preparing, this is to be applied AFTER #626 

- [x] Closes #635

### DIFF
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -260,6 +260,7 @@ class TCPIPInstrHiSLIP(Session):
             lock_timeout = calculate_lock_timeout_from_open_timeout(self.open_timeout)
             _k, rv = self.lock(constants.Lock.exclusive, lock_timeout, "")
             if rv != StatusCode.success:
+                self.close()  # the operation was open_with_lock. If the lock fails, the open should also be reversed.
                 raise RuntimeError("Failed to acquire exclusive lock")
 
     def _handle_async_service_request(self, status_byte: int) -> None:
--- a/pyvisa_py/tcpip.py
+++ b/pyvisa_py/tcpip.py
@@ -114,7 +114,7 @@ class TCPIPInstrHiSLIP(Session):
     session_type = (constants.InterfaceType.tcpip, "INSTR")
 
     _supported_event_types = {constants.EventType.service_request}
-    
+
     REMOTELOCALOPCODE: Final[dict[constants.RENLineOperation, str]] = {
         constants.RENLineOperation.address_gtl: "justGTL",
         constants.RENLineOperation.asrt: "enableRemote",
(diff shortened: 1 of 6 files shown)


hb020:
HiSLIP transforms into open-with-lock into open-and-then-lock.

In view of the problems with DMM6500, we should maybe also move VXI-11 to that method, just like NI-VISA does it.

pyvisa_py/tcpip.py
```diff
@@ -230,6 +250,19 @@ def after_parsing(self) -> None:
...
+            # the interface likely supports it, but for now we only handle exclusive locks.
+
+            # about the timeout: Found nothing in the spec that says how to handle this.
+            # Do what is done by VXI-11
+            lock_timeout = calculate_lock_timeout_from_open_timeout(self.open_timeout)
+            _k, rv = self.lock(constants.Lock.exclusive, lock_timeout, "")
+            if rv != StatusCode.success:
+                self.close()  # the operation was open_with_lock. If the lock fails, the open should also be reversed.
+                raise RuntimeError("Failed to acquire exclusive lock")
+
```
MatthieuDartiailh: Shouldn't the lock attribute be set now.

pyvisa_py/tcpip.py
```diff
@@ -203,7 +222,8 @@ def after_parsing(self) -> None:
         self.attrs[ResourceAttribute.read_buffer_operation_mode] = (
             constants.VI_FLUSH_DISABLE
         )
-        self.attrs[ResourceAttribute.resource_lock_state] = constants.VI_NO_LOCK
+        # do NOT set the resource lock state manually; it will be managed by the remote locking mechanism.
+        # self.attrs[ResourceAttribute.resource_lock_state] = constants.VI_NO_LOCK
```
MatthieuDartiailh: Remove the comment

MatthieuDartiailh (commented):
Looks good. One comment can be removed.

Did I get that right that one cannot query the lock state under VXI-11 ?

pyvisa_py/tcpip.py
hb020: removed the ` # self.attrs[ResourceAttribute.resource_lock_state] = constants.VI_NO_LOCK`

hb020:
> Did I get that right that one cannot query the lock state under VXI-11 ?

You can query PyVISA-Py, but not the instrument (unlike HiSLIP).

MatthieuDartiailh (approved):

